PR #19, opened 2021-09-16
some fixes , primarly targetting macos

usama-makhzoum:
fixed creating profile with firefox.app, and shell quoting in general


hassandraga:
The auto test (_MinGW-Windows_) and (_MSVC-Windows_) both success. Thank you for this PR.

PR #25, opened 2022-10-28
Disable First Run Dialog

malisipi:
This pr disables chrome's first-run dialog at new profile creation.
chrome-welcome


AlbertShown:
This will improve the UX, thanks.

PR #57, opened 2023-03-18
Update webui.c

CardealRusso:


AlbertShown:
Thank you for your contribution. I test it, and it's working fine. But unfortunately, your changes are mixing Chrome & Chromium.

```c
typedef struct webui_browser_t {
    unsigned int chrome;    // 1
    ...
    unsigned int chromium;  // 5
} webui_browser_t;
```

So instead, we should create a new section for Chromium:

1. Adding `else if(browser == webui.browser.chromium)` in `_webui_browser_exist()`.
2. Clone `_webui_browser_start_chrome()` and rename it to `_webui_browser_start_chromium()` with the necessary functionality changes.
3. Create a new PR

CardealRusso:
chrome and chromium are the same, only the icon and the name change. unlike edge, which is a totally different fork (with microsoft's own spyware) I honestly don't see plausible reasons to separate chrome and chromium.

AlbertShown:
You are totally right. Indeed, there is no base code difference between Chrome and Chromium except the Google proprietary code (_AddOns platform, MP3, H. 264, AAC codecs..._). But I remembered last year, a user wanted to run his apps on Chromium and not touch Google Chrome, which made me add Chromium as a separate option.

But of course, `webui_show(my_window, "", webui.browser.any);` will try to run Chrome. If not, then Chromium.

PR #63, opened 2023-03-25
Add Chromium support

CardealRusso:
Not sure if it is working

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@
 ## Features
 
 - Written in Pure C
-- Fully Independent (*No need for any third-party library*)
 - Lightweight (*~160 Kb using TCC*) & Small memory footprint
 - Fast WS binary communication (*App--WebUI--Browser*)
 - One header file
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
 ## Features
 
 - Written in Pure C
+- Fully Independent (*No need for any third-party library*)
 - Lightweight (*~160 Kb using TCC*) & Small memory footprint
 - Fast WS binary communication (*App--WebUI--Browser*)
 - One header file
--- a/src/webui.c
+++ b/src/webui.c
@@ -2372,9 +2372,9 @@ bool _webui_browser_start(webui_window_t* win, const char* address, unsigned int
             return _webui_browser_start_edge(win, address);
         else if(browser == webui.browser.custom)
             return _webui_browser_start_custom(win, address);
-				else if(browser == webui.browser.chromium)
-						return _webui_browser_start_chromium(win, address);
-				else
+        else if(browser == webui.browser.chromium)
+            return _webui_browser_start_chromium(win, address);
+        else
             return false;
     }
     else if(win->core.CurrentBrowser != 0) {
@@ -2388,9 +2388,9 @@ bool _webui_browser_start(webui_window_t* win, const char* address, unsigned int
             return _webui_browser_start_edge(win, address);
         else if(win->core.CurrentBrowser == webui.browser.custom)
             return _webui_browser_start_custom(win, address);
-				else if(browser == webui.browser.chromium)
-						return _webui_browser_start_chromium(win, address);
-				else
+        else if(browser == webui.browser.chromium)
+            return _webui_browser_start_chromium(win, address);
+        else
             return false;
             //webui::exit();
     }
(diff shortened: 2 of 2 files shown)


hassandraga:
This PR seems legit, Thank you for your contribution.

But the commit `It depends on mongoose` need to be removed first as the end-user can clone WebUI and start using it **without the need** to install any other library or runtimes.

WebUI 1.0 was written in C++ and it need many Boost dll to run, so v1.x it wasn't fully independent. But 2.x is independent.

CardealRusso:
I'm not using github desktop, I can't revert a commit

hassandraga:
LGTM.

CardealRusso:
it is not working and i dont know why

hassandraga:
Can you please share the logs? Rebuild the library in the debug mode `make debug` or `mingw32-make debug`.

CardealRusso:
> Can you please share the logs? Rebuild the library in the debug mode `make debug` or `mingw32-make debug`.

for some reason it is trying to find chromium on wrong directory

```[0] _webui_file_exist([C:\Program Files\Chromium\Application\chrome.exe])...
[0] _webui_file_exist([C:\Program Files (x86)\Chromium\Application\chrome.exe])...
[0] _webui_free_port([10736])...
[1] webui_show(html, [0])```

**Comment by hassandraga** [comment]
> Does the "%s%s" on code means the program files folder?

Yes. 

So, this is the challenge: the function `_webui_browser_exist()` should check whether a specific web browser is installed. On Linux, it's easy, we just run `chrome --version` and check for the exit code, but in Windows, it is different. Usually, we can use `start /wait chrome --version`, but this will show an error message box if the web browser is not installed.

image

The final solution we did in Windows is to check if the file exists in `Program Files`. It's working, but the issue is that the browser can be installed anywhere else, making WebUI fail to find it.

I'm planning when I have time to update `_webui_browser_exist()` to check the registry keys, but if anyone has a better idea, please let us know.

**Comment by CardealRusso** [comment]
On BrowserAny on windows, you could get the path to the default browser (html handler) on reg
```
HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\FileExts\.html\UserChoice
```

then 
```
HKCR\(ProgId)\shell\open\command
```

After that, you just need to  find a way to assume which browser it is (whether chromium based or firefox)

CardealRusso:
> > Does the "%s%s" on code means the program files folder?
> 
> Yes.
> 
> So, this is the challenge: the function `_webui_browser_exist()` should check whether a specific web browser is installed. On Linux, it's easy, we just run `chrome --version` and check for the exit code, but in Windows, it is different. Usually, we can use `start /wait chrome --version`, but this will show an error message box if the web browser is not installed.
> 
> image
> 
> The final solution we did in Windows is to check if the file exists in `Program Files`. It's working, but the issue is that the browser can be installed anywhere else, making WebUI fail to find it.
> 
> I'm planning when I have time to update `_webui_browser_exist()` to check the registry keys, but if anyone has a better idea, please let us know.

Plus, if you call for the program without start, it wont return a message box

hassandraga:
> HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\FileExts\.html\UserChoice

This is a good start. We may need to find the registry key for each web browser.

> Plus, if you call for the program without start, it wont return a message box

I'm sure the web browser's setup does not add it to the system path. The only way to start the browser in Windows is by using `start`.

AlbertShown:
> HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe
> HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe

I found no web browser in `HKEY_CURRENT_USER`, but I found Chrome, Firefox, and Edge in `HKEY_LOCAL_MACHINE`.

AlbertShown:
Where did you download the Chromium installer?

AlbertShown:
I got it now. https://learn.microsoft.com/en-gb/windows/win32/shell/app-registration#using-the-app-paths-subkey

> An application that is installed for per user can be registered under HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\App Paths. An application that is installed for all users of the computer can be registered under HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\App Paths.

CardealRusso:
This solution is perfect, at the cost of what i already said on #57 to @AlbertShown :
It is very redundant and unnecessary to separate chrome from chromium.

And in this solution, the problem gets even worse, because both chrome and chromium use the same executable name "chrome.exe", so this solution will just "assume" if it is chrome or chromium through the directory path, which is a very bad practice because the user may be running chrome/chromium in a custom directory.

I consider this solution to be perfect also because of custom directories possibility (portable installations).

CardealRusso:
https://ungoogled-software.github.io/ungoogled-chromium-binaries/

AlbertShown:
You are right. Both Chrome and Chromium have the `chrome.exe` key. On top of that, I tested all browsers, and all Chromium-based web browsers can open the window using the same profile folder.

**Registry keys**:

**Chromium**: `chrome.exe` -> Default Key -> `C:\...\Chromium\Application\chrome.exe`
**Chrome**: `chrome.exe` -> Default Key -> `C:\...\Google\Chrome\Application\chrome.exe`
**Edge**: `msedge.exe` -> Default Key -> `C:\...\Microsoft\Edge\Application\msedge.exe`
**Epic**: `epic.exe` -> Default Key -> `C:\...\Epic Privacy Browser\Application\epic.exe`
**Yandex**: `browser.exe` -> Default Key -> `C:\...\Yandex\YandexBrowser\Application\browser.exe`
**Vivaldi**: `vivaldi.exe` -> Default Key -> `C:\...\Vivaldi\Application\vivaldi.exe`
**Brave**: `brave.exe` -> Default Key -> `C:\...\BraveSoftware\Brave-Browser\Application\brave.exe`
**Opera**: `opera.exe` -> Default Key -> `"C:\...\Opera\Launcher.exe"`
**Firefox**: `firefox.exe` -> Default Key -> `C:\...\Mozilla Firefox\firefox.exe`

I agree with what @CardealRusso proposed, which is combining Chrome & Chromium.

And I suggest removing the param `browser` from `webui_show(webui_window_t* win, const char* html, unsigned int browser)`  so the developer won't worry about the browser, and WebUI will try to read registry keys and search in this order:

**Top Prefered**
1. Chrome - Because it works perfectly
2. Edge - Because it works perfectly like Chrome
3. Epid - Because it works perfectly like Chrome
4. Vivaldi - Because it works perfectly like Chrome

**Minor Issues**
1. Brave - Shows a policy notification in the first run
2. Firefox - Does not support App-Mode like Chrome (_Looks not great_)
3. Yandex - Shows a big welcome window in the first run
4. Chromium - Some Anti-Malware shows a false alert when using [ungoogled-chromium-binaries](https://ungoogled-software.github.io/ungoogled-chromium-binaries/)

AlbertShown:
Opera does not work at all !

I will implement all those suggestions.
Thank you @CardealRusso, for your Registry implementation code.

Issue #56, opened 2023-03-17
it wont detect ungoogled-chromium


CardealRusso:

CardealRusso:
it seems like chromium dont work at all on windows

neroist:
It seems like it doesn't, even when Chromium is added to path... In fact the string "chromium" is only found in the source code once.

hassandraga:
I tested Chromium on Linux a while ago, but not on Windows. I will look into this today.

Thanks for reporting.

PR #64, opened 2023-03-26
Fix Chromium Path

CardealRusso:
Sorry for not testing before. I compiled this patch and it is working nicely
Sorry, im not familiar with C
hTdKjjq 1


CardealRusso:
This patch also definitively closes #56

hassandraga:
LGTM. Thank you @CardealRusso.

PR #92, opened 2023-05-03
Fix #90: Support 'chromium'

### LINKED ISSUE #90: Chromium should be used on Linux
Some distros ship chromium (FOSS Chrome build) under binary name `chromium`. I don't know much C, but it seems that currently webui searches for chromium only on macOS. Since Chromium/chrome has better app mode, it would be nice to try to use it on Linux as well.

malisipi:


AlbertShown:
LGTM. Thank you 👍

Issue #103, opened 2023-05-21
No specified C++ standard in make files and potential issues


GiuseppeCesarano:
The make files responsible for compiling the C++ examples in the project currently lack a selected C++ standard, and there is no mention of the targeted C++ standard within the project.

This could lead to:  

1) Lack of feature awareness for contributors  
Contributors may face difficulties in determining which C++ language features are available to them. For instance, the current implementation of the `window::return_string` member function in the C++ bindings:  
```c++
void return_string(webui::event* e, std::string s) {
  webui_event_t* c_e = convert_event_to_webui_event(e);
  webui_return_string(c_e, &s[0]);
  delete c_e;
}
```
This demonstrates passing a string by value, which can be computationally expensive. However, in this particular case, such copying is unnecessary, as the function `webui_return_string` already creates a copy of the string. To mitigate this, there are two main approaches: passing the `std::string` as `const std::string&` or utilizing `std::string_view`. The latter is the recommended approach due to its ability to alias to various types and potentially avoid costly calls to the `std::string` constructor. Nevertheless, adopting `std::string_view` requires C++20, and implementing this change with the current make files will result in compilation errors.

2) Inconsistency across compilers  
By not explicitly defining the C++ standard, the compilers will resort to defaulting to a particular standard based on the compiler version. For instance, clang 15 will default to using C++14, while clang 16 will default to C++17. This inconsistency can lead to the infamous "it works on my PC" problem.

Additionally, considering the nature of the project and the potential benefits of modern C++ features, it is recommended to adopt at least C++20 as the targeted standard within the make files.

hassandraga:
Thank you, Giuseppe, for pointing this out.

Honestly, this project is purely C99, but someone on Reddit requested a C++ wrapper, so I created the C++ wrapper in 20-25 minutes in a rush... and for sure, it's not complete and needs improvements.

You are right about the unnecessary strings copy, so I update it to use references instead of values, especially when APIs take `const char*`. So, thank you for reporting that.

About the standard, the WebUI C++ wrapper is relatively small and easy, so selecting a minimum standard as C++11 is enough. This helps developers who are using WebUI in old embedded/PC devices. On the other hand, anyone can easily modify the `webui.hpp` and `makefile` to set a different standard as needed.

Fix: https://github.com/webui-dev/webui/commit/cdf67b17ac65c6467733fbd0adcbfe17f084ac02

GiuseppeCesarano:
[...]

PR #105, opened 2023-05-23
Adherence to noninvasive C++ best practices

GiuseppeCesarano:
# Changes details

## nullptr instead of NULL
`NULL` could be implicitly cast to `int(0)`, `nullptr` prevents this and has no downside.

## Const-ed member function where possible
Simple quality of life change, it has no downside and allows users to call the const-ed function even if it has a const instance.

## Empty window destructor removed.
Empty destructors and explicit default constructors prevent the compiler from automatically generating copy constructors, move constructors, and assign operators; since it makes the compiler think that the specific class has a special lifetime, this isn't the case for the `window` class, so removing the destructor results in its automatic generation and the automatic generation of the other functions. 

## Using member initializer in constructor
This prevents members from being initialized by default and then copied.

## Using std::string_view instead of const std::string&.
See #103

## Using inline with free functions
This prevents the ODR rule violation
_Note: This is not function inlining that prevents stack push and stack pop._

This changes are tested on Linux using Firefox as browser and Clang and GCC as compilers.

### DIFF
--- a/include/webui.hpp
+++ b/include/webui.hpp
@@ -13,6 +13,7 @@
 
 // C++ STD
 #include <string>
+#include <string_view>
 
 // WebUI C Header
 extern "C" {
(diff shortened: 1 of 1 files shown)


hassandraga:
Tested on Windows using GCC `v12.2.0` and MSVC `v19.35`. Works fine.
Thank you for this improvement. I appreciate it.

PR #106, opened 2023-05-24
C++ header restructured.

GiuseppeCesarano:
# C++ header restructured.

## Goals of this PR
- Eliminate the need for a conversion function between webui::event and webui_event_t.
- Incapsulate the event handling capabilities in a static class.
- Break the circular dependency between types defined in the C++ header.

## Non-goals of this PR
This PR is only intended to be a code restructuring, and doesn't mean to change how the header achieves the functionality it provides.

## Notes
- These changes slightly change the API, so the example needs to be updated (this PR encapsulates the event struct inside the window class, so to declare an event instance you now need `webui::window::event` instead of `webui::event`).
- This PR is marked as draft because I expect a discussion on these changes, if the changes are well received, I will update the examples.
- These changes have already been tested on Linux using firefox as browser, clang and gcc as compiler, but only with the `serve_folder` example

### Eliminate the need for a conversion function between webui::event and webui_event_t.
This is accomplished by making the C++ event struct a derived class of the C struct;
This allows us to use object slicing and pass the C++ struct pointer directly to the C API.
The proposed implementation of the C++ class has two "references" to the window object, one is the `const webui::window&` member, the other is the `webui_event_t.window` member; if the user misuses the API, these two could fall out of sync, turning the `event struct' into a class with the proper gardrails could be the best solution, but is a bit more invasive since it would change the API, so this change isn't currently in this PR.

### Incapsulate the event handling capabilities in a static class.
This change is achieved by encapsulating all free functions and global variables in a static class called handler. There is no particular downside to this.

### Break the circular dependency between types defined in the C++ header.

Once the class that handles the events was created, the circular dependency between the types became apparent, making the implementation literally impossible. 

So the proposed solution is a stronger gerarchy of types. Now the window type encapsulates the event type, which encapsulates the handler type.

As a result of these changes and the removal of the conversion function, the window class had some static members that accepted an `event*` like `get_string(event*)` and `get_bool(event*)`; these members have been moved to the event subclass.

### DIFF
--- a/examples/C++/call_cpp_from_js/main.cpp
+++ b/examples/C++/call_cpp_from_js/main.cpp
@@ -6,12 +6,12 @@
 // Include C++ STD
 #include <iostream>
 
-void my_function_string(webui::event* e) {
+void my_function_string(webui::window::event* e) {
 
     // JavaScript:
     // webui_fn('MyID_One', 'Hello');
 
-    std::string str = e->window.get_string(e);
+    std::string str = e->get_string();
     std::cout << "my_function_string: " << str << std::endl; // Hello
 
     // Need Multiple Arguments?
@@ -25,38 +25,38 @@ void my_function_string(webui::event* e) {
     // bar = my_json[1];
 }
 
-void my_function_integer(webui::event* e) {
+void my_function_integer(webui::window::event* e) {
 
     // JavaScript:
     // webui_fn('MyID_Two', 123456789);
 
-    long long number = e->window.get_int(e);
+    long long number = e->get_int();
     std::cout << "my_function_integer: " << number << std::endl; // 123456789
 }
 
-void my_function_boolean(webui::event* e) {
+void my_function_boolean(webui::window::event* e) {
 
     // JavaScript:
     // webui_fn('MyID_Three', true);
 
-    bool status = e->window.get_bool(e); // True
+    bool status = e->get_bool(); // True
     if(status)
         std::cout << "my_function_boolean: True" << std::endl;
     else
         std::cout << "my_function_boolean: False" << std::endl;
 }
 
-void my_function_with_response(webui::event* e) {
+void my_function_with_response(webui::window::event* e) {
 
     // JavaScript:
     // const result = webui_fn('MyID_Four', number);
 
-    long long number = e->window.get_int(e);
+    long long number = e->get_int();
     number = number * 2;
     std::cout << "my_function_with_response: " << number << std::endl;
 
     // Send back the response to JavaScript
-    e->window.return_int(e, number);
+    e->return_int(number);
 }
 
 int main() {
--- a/examples/C++/call_js_from_cpp/main.cpp
+++ b/examples/C++/call_js_from_cpp/main.cpp
@@ -9,13 +9,13 @@
 #include <string>
 #include <stdexcept>
 
-void my_function_exit(webui::event* e) {
+void my_function_exit(webui::window::event* e) {
 
     // Close all opened windows
     webui::exit();
 }
 
-void my_function_count(webui::event* e) {
+void my_function_count(webui::window::event* e) {
 
     // This function gets called every time the user clicks on "MyButton1"
 
--- a/examples/C++/serve_a_folder/main.cpp
+++ b/examples/C++/serve_a_folder/main.cpp
@@ -13,7 +13,7 @@ class MyClass {
 
     // This method gets called every time the
     // user clicks on "OpenNewWindow"
-    void show_second_window(webui::event* e) {
+    void show_second_window(webui::window::event* e) {
 
         // Show a new window, and navigate to `/second.html`
         // if the window is already opened, then switch in the same window
@@ -22,7 +22,7 @@ class MyClass {
 
     // This method gets called every time the
     // user clicks on "SwitchToSecondPage"
-    void switch_to_second_page(webui::event* e) {
+    void switch_to_second_page(webui::window::event* e) {
 
         // Switch to `/second.html` in the same opened window.
         e->window.show("second.html");
@@ -31,7 +31,7 @@ class MyClass {
     // Example of a simple function (Not a method)
     // This function receives all events because
     // it's get bind with an empty HTML ID.
-    void events(webui::event* e) {
+    void events(webui::window::event* e) {
 
         if (e->event_type == webui::CONNECTED)
             std::cout << "Window Connected." << std::endl;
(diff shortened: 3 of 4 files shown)


GiuseppeCesarano:
> I was initially confused about get_string()

I was also unsure about moving members like this into the `event` struct, but leaving them in the window `class` as static that act on a `event` was too weird.

> I don't mind implementing webui_event_t.window.

I don't really understand what you mean with that; with the current implementation the `webui_event_t.window` is already present in the C++ version of `event` and so is a const reference to the window object; these two things should always "refer" to the same window, but as the current implementation stends they could fall out of sync (if the user messes with it), basically what should be implemented are getters and setters (we should also add them for the other members) thus the users can't make these two variables out of sync.

> About the static handle member, the code is clear and compiles fine. Still, I'm wondering if we have, let's say, 10 opened windows, and every window use a JavaScript that call webui_fn() 10 times per second. that means we have 10 x 10 events per second. In this case, having one static instance of handle is better or having multiple (one for every window)?

The proposed implementation does not change this aspect of the hpp file, since the way the static class `hendler' is compiled is equivalent to a namespace with a "private" section where its members are namespace-wide (as it was before, except for the private part).

So if there's any doubt about this particular implementation, there's not much to worry about, since it's equivalent as before.

If the doubts are extended to the original implementation, I would speculate that having more than one array wouldn't change much, I don't think any race conditions occur in the current implementation and anyway the callbacks array and the windows array once created are basically only read. As for cache performance, each CPU should be able to have a copy in the cache if the compile decides to go that way.

AlbertShown:
Yes, that's what I was wondering about, if there are no race conditions then your proposal is better than the original.

GiuseppeCesarano:
Perfect! Tomorrow I will push a commit that will update all the C++ examples.  

For now, dead giveaways that there aren't really any instances of the `handler` class are that there are no constructors or destructors, since they are explicitly deleted, and the `handler::handle` function has the same signature as a free function (so we can pass it to `webui_bind`).

GiuseppeCesarano:
All examples have been updated except `VS2019` and `VS2020` which use a locally defined `webui.hpp`.

As a result of the member function changes, the API looks cleaner in the examples, but I'm not 100% happy with the `event' struct, I feel like it should be a class with it's own methods that prevent the user from messing up the internal members.

GiuseppeCesarano:
Making these updates to the examples made me realize that the `window&` inside the `event` struct isn't needed anymore.

This is a good thing, as it makes the C++ event literally the same type as the C event, with the addition of some member functions.

This also fixes the synchronization problem of the two variables referring to the same thing in two different ways; however, I still think accessors are needed for the `event` class, because as it is now, directly accessing it's members would still feel a bit C-like.

But I think the API design for the `event` struct is outside the scope of this PR, and should be discussed in it's own thread.

hassandraga:
> All examples have been updated except VS2019 and VS2020

The `webui.hpp` files inside VS2019 and VS2020 are copies of the original `/include/webui.hpp`. You can manually copy it, or run `linux_build.sh` to copy all headers and build the C library.

> made me realize that the window& inside the event struct aren't needed anymore

I'm confused because the event should have a ref to the window object where the event happened. Let's say a user used `bind("")`  from multiple windows to the same callback, which makes the callback receives all events from all opened windows. A ref to that window is needed. This will make calling `script()` from an event possible.

This commit is broken: https://github.com/webui-dev/webui/commit/fa0918e862f630d3644577b7913542862d033b83
`e->window.script` / `e->script` --> event cannot call any window class members.

Please correct me if I'm wrong.

GiuseppeCesarano:
The window reference isn't strictly needed anymore, since the new event struct already has the C way of distinguishing which window generated the event.

Of course, this is bad because it forces the user to use the C API; but it's easy to fix with two functions.

in `class handler`
```c++
static webui::window& get_window(const size_t index){
    return *window_list[index];
}
```

in `struct event`
```c++
webui::window& get_window(){
    return event::handler::get_window(window);
}
``` 

If needed I can open a another pull request where those two functions are added and the examples are fixed.

GiuseppeCesarano:
Also, what kind of formatting tool are you using? I have clang-format, but it looks like it can't pick up a configuration file

AlbertShown:
That's mean `e->window.script()` will become:

```
webui::window MyWindow = webui::get_window(e->window);
MyWindow.script()
```

Many developers prefer a simple API library usage more than knowing that the library is very efficient behind the scenes. I mean, `e->window.script()` was so simple to use, even if the implementation behind the scenes was not the best.

GiuseppeCesarano:
That's not how the call would work, as mentioned before, the `get_window()` function would be part of the `event` struct.
So the old `e->window.script()` should be rewritten as `e->get_window().script()`.

I'm opening another PR to implement the change.

AlbertShown:
The implementation is working fine. But what about making the APIs simple as possible?
Why not we change back `struct event : public webui_event_t` to a C++ struct that has a window object `e->window`?
The cost will be creating a new C++ event struct every time... it's fine for a simple click but bad for scripts that send hundreds of events every second... 🤔

It's hard to decide, so let's keep your implementation because it's fast and lightweight 👍

GiuseppeCesarano:
> The implementation is working fine. But what about making the APIs simple as possible?
Why not we change back struct event : public webui_event_t to a C++ struct that has a window object e->window?

I strongly belive that the current API is cleaner and simpler. Some reasons are listed below:
- Adding back a window reference would mean that the window object is (again) "pointed to" by two separate variables, the `window&` and the `window_event_t.window` (which is created with the base of the C++ `event`); now that we have two separate variables that should do the same thing, how do we deal with that? Which one is the "trusted" one?  Do we keep them synchronized? If not, and if we only "trust" and use one of the variables, why does the other one exist?
- Having a function call instead of accessing the members directly doesn't allow the user to misuse the variable.
- A C++ user expects a method for such things, it's kinda unusual to access members directly (for this reason, and for API consistency, the other members should also be added, and the struct `event` should become a class).

Ignoring all that, is `e->window` so much better than `e->get_window()`?

> The cost will be creating a new C++ event struct every time... it's fine for a simple click but bad for scripts that send hundreds of events every second... thinking

This would only happen if we completely revert the PR, as for now if we add the `window&` the cost should only be the population and managing it's synchronization (or lack thereof). But as stated before I don't see value in that

AlbertShown:
You have a point 👍
After reviewing the changes and your explanation, I saw that your improvement is way better than the original. So, thank you @GiuseppeCesarano, for that.

hassandraga:
Thank you @GiuseppeCesarano for your C++ contribution. I like it.

Issue #104, opened 2023-05-23
Vivaldi for Arch based distros


demetera:
Vivaldi binary in Arch / Manjaro distros called by `vivaldi-stable` while there's no checks for this behaviour in the code:

```
Line 2621:
if(_webui_cmd_sync(win, "vivaldi --version", false) == 0) {
```
```
Line 2623:
sprintf(win->browser_path, "vivaldi");
```

![image](https://github.com/webui-dev/webui/assets/3475944/2c3d2095-731b-44fc-8389-8a9f0e897479)

Thanks!

hassandraga:
Thank you for reporting this. I have `vivaldi` in my Ubuntu setup, so, would you like please duplicate the line and add `vivaldi-stable`, then test it and create a PR. Otherwise, I can create a PR and you test it later in your environment.

Dialga:
There is also `vivaldi-snapshot`, can another line be added to test for that?

AlbertShown:
If `vivaldi-snapshot` can exist on the end-user's machine and it's not an exception, then yes. Please add a PR.

demetera:
Thank you for merging `vivaldi-stable`
Made the same for `vivaldi-snapshot`
meanwhile unable to test at the moment

demetera:
Thanks for merging!

I had an idea  for these kind of builds, like canary, devs, stables, snapshots, betas etc. (every chromium based browser applies though), just to add a clause on main README.md with recommendation to create something like `alias vivaldi="vivaldi-snapshot"` in bash configuration file

AlbertShown:
Good idea. Perhaps we can just hardcode them by priority.
 Can you please create a list by priority?

**I'm thinking:**
```
vivaldi
vivaldi-stable
vivaldi-canary
vivaldi-dev
vivaldi-snapshot
vivaldi-beta
```

Then we apply all of them to all Chromium-based browsers.

Dialga:
> Good idea. Perhaps we can just hardcode them by priority. Can you please create a list by priority?
> 
> **I'm thinking:**
> 
> ```
> vivaldi
> vivaldi-stable
> vivaldi-canary
> vivaldi-dev
> vivaldi-snapshot
> vivaldi-beta
> ```
> 
> Then we apply all of them to all Chromium-based browsers.

Brilliant, although for opera they have `opera-developer`.

demetera:
Dears,

this is the list of binaries for Chromium based browsers, its repository and if it's AUR package - number of votes:

[...]

PR #107, opened 2023-05-24
vivaldi fix for Arch

demetera:
Fix for the issue #104 

Added condition to check `vivaldi-stable` on Arch based distros additinally to `vivaldi`
Tested on Arch and Ubuntu.


hassandraga:
Thank you for your contribution.

PR #111, opened 2023-05-26
Vivaldi-snapshot condition for Arch

demetera:
#104 
Currently I can't download & make vivaldi-snapshot from AUR  to test, but this should work


AlbertShown:
Yes, It should work if the snapshot version is installed.
Thank you for your contribution.

PR #113, opened 2023-05-27
C++ header reorganization completed.

GiuseppeCesarano:
As the title says, this is the final PR for the C++ header reorganization; now the C++ API should be in a good place, being as fast as the C with OOP in mind.

Main changes in this PR:
- Now event is a class with it's own methods (see [this](https://github.com/webui-dev/webui/pull/106#issuecomment-1563526571))
- Some examples still used `e->window` instead of `e->get_window()` 

All examples are tested and confirmed to work on Linux with Firefox as browser and clang or gcc as compiler.


hassandraga:
Thank you for this series of improvements. I like the implementation you did.

PR #129, opened 2023-07-06
fix: html error pages displayed as plain text

JOTSR:
# Issue
Send `webui_html_*` erros via `mg_send_http_*`  use the wrong mime type (text/plain) and so badly display the error.
image

# Fix
Civetweb only allow serving error page as text/plain so need I added a new webui internal function to serve text/html error pages.
image


AlbertShown:
Me too. I noticed this issue a while ago but didn't report it!
Thank you @JOTSR 👍

PR #130, opened 2023-07-06
fix: :bug: webui js API is undefined

JOTSR:
# Issue

image
image

Since webui script tag is injected after html, any declared js file in the html payload is parsed before the API.
Unless set it "defer" it is executed before the API and cause an error.

I change the html structure order to force js bridge to execute before any user content.
Since bridge is loaded in html head I fix possibly undefined listener that crash the app.

Moreover, since webui.c allow to load js via 2 way (embbeded and via a get url) I choose to the the url way to avoid memory management for future improve and simplify the code (and js was truncated after the changes if I keep string concat approach). In bonus it allow to debug js via the source panel in dev console.

# Improvements

To force top loading I used the auto correct behaviour of html and then construct bad html to load js bridge before all user content (script is loaded just afet <html> tag that is never close, any added user content will be merge after and js result to be in <head>).
This is more an hot fix and need to be improve by inserting the script tag in html head via an XML/DOM parser.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -294,11 +294,11 @@ static const char* webui_javascript_bridge =
 "        _webui_send_event_navigation(url); \n"
 "    }); \n"
 "} \n"
-"document.body.addEventListener('contextmenu', function(event){ event.preventDefault(); }); \n"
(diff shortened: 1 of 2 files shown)


neroist:
LGTM

PR #131, opened 2023-07-06
refactor: split js bridge source code from lib source code

JOTSR:
# Motivation

Embedding js bridge source code into c make it hard to read and edit. Moving it appart was made possible by using xxd tool.
- It open new possibility like using typescript for API and transpile to js for dist.
- It improve dev experience and readability.
- Allow linter and synthax analyzer to check the code.
- Allow minification in order to reduce lib size.
- Allow user to import type definition to better discoverability and usage of the API.

# Possible improvements
- Move error pages, svg and all big embedded content to specific file for better readability.
- Set up minification.
- Refactor js api.
- Mirror type definition on accessible url (for instance deno.land can serve a subfolder as module and track the main project version as module specifier).


AlbertShown:
The idea of spitting the main code into multiple files is a great idea, but it comes with a cost.
Many users won't have `xxd` installed. Probably we should use a Python script to convert `JS` to `C-Style`, but I'm not sure.
Another option is using C23 Standard to use `#embed` to automatically embed JS file. But that means removing the TCC support.

I don't know what we should do honestly, but yes, splitting files is a good improvement.

JOTSR:
I'm not sure that xxd is not installed. It seem's to be preinstalled on many linux distrib and came natevely on macos. It is also a part of git for windows (part of git bash tools).

I actually working on the js bridge (there is some errors, deprecated APIs, overcomplicated code, ...). And to simplify its maintenance and its integration for user I asking if typescript would be a good idea. So for compiling ts I propose to setup a **devcontainer**, it can be useful to ensure all dev have all necessary tools. As there is actually no linting or code formatter in the project, a devcontainer with adequat tools can be a good idea to homogenize and support development.

I don't known how the support of TCC matter (I'm not a C dev so I am not aware of current usage) but `#embed` is a interesting feature.

AlbertShown:
> typescript would be a good idea

The Deno wrapper is written in TypeScript. So, no one will have an issue switching the WebUI-Bridge from JavaScript to TypeScript.

> I propose to setup a devcontainer

I vote for the devcontainer to be optional, as many developers prefer to edit `/client/webui.ts` directly. But it's good to have containers for WebUI dev.

> I don't known how the support of TCC matter

TCC is for users who want to produce a complete working GUI executable using WebUI in less than 1024 Kb. Also, TCC is used as a backend in V-WebUI.

JOTSR:
# Bridge
> The Deno wrapper is written in TypeScript. So, no one will have an issue switching the WebUI-Bridge from JavaScript to TypeScript.

The JS bridge inside the c lib is totally independent from the Deno wrapper. For instance if you use TSX for your UI, as the client API is not typed it lack of usability.

Using typescript (or at least providing a d.ts definition file) can be a good improvement for discoverability, interoperability and simplify project integration. As it only used in source code it don't change the final execution (only js i embedded).

Moreover, using typescript bring checking tools for internal use (there is several issues in the bridge that are catched by typescript in my tests).

# Devcontainer
Yes .devcontainer is only for simplify startup and project tools sharing (not mandatory). For instance there is no default formater or linter for the project and requiring from contributors to install project choosen tools can be complicated. Future CI/build improvements can take profit of it.

AlbertShown:
> The JS bridge inside the c lib is totally independent from the Deno wrapper.

Yes, I meant that the original creator of WebUI already used TypeScript, so I don't see any issue in switching the bridge completely from JavaScript to TypeScript. It should just simply work in the browsers.

> Bridge
> Devcontainer

I totally agree with you. I guess this will be a very good improvement 👍

Issue #98, opened 2023-05-08
Add window size, position


kbkpbot:
something like that 
```c
  webui_set_window_title(window, 'title');
  webui_set_window_size(window, '800x600');
  webui_set_window_position(window,'100,100');
```

hassandraga:
I guess yes, `webui_set_window_size` and `webui_set_window_position` will be helpful. We will add it to the command line.
While `webui_set_window_title` is a user's responsibility to do in the HTML `<title>My Title</title>`.

malisipi:
Also `window.resize();` and `window.move();` can be useful to move/resize after creating window.

AlbertShown:
I was thinking last week about how to use Win32 APIs, GTK APIs, and Cocoa APIs to change the window position and size. Now you are telling me we can do this easily using JavaScript? I'm dump 🤦‍♂️

hassandraga:
@HoffmannP As you said, we can change and close the window using JavaScript because of how the window is opened. Let's implement this.

wwderw:
I don't know if its a Firefox issue or what, but I made sure that there wasn't anything turned off that may have kept the browser for respecting that, but by default everything appeared to be turned off that I was looking at.

I was unable to get window.resizeTo(x,y) to work directly in the HTML document or on the C++ side injecting with run() after made sure that the window was up and running.  Is there something that I'm missing or has it changed to where that isn't respected, at least within Firefox?

malisipi:
@wwderw Can you give me some details about OS, firefox version and chrome support. If you're using Linux, which display server are you using? (If you're not using Linux, just ignore the last question)

wwderw:
OS: Manjaro
Browser Version: 114.0.1
Display: X11

> @wwderw Can you give me some details about OS, firefox version and chrome support. If you're using Linux, which display server are you using? (If you're not using Linux, just ignore the last question)

petabyt:
Setting the window size should be easily possible, see #1. It it hasn't been implemented yet, I can send in a PR. I need it for my application. I was able to do it with JS, but that causes the window to jump (starts out as a large window before being resized a few ms after opening)

AlbertShown:
Yes, @petabyt, You can create a PR, but before, write a pseudo C code to how this new API will look like.

Suggestion:
```c
webui_set_window_size(1024, 1024);
webui_set_window_position(512, 512);

webui_show()...
```

petabyt:
```
[...]

PR #132, opened 2023-07-12
refactor, fix and add feat to js bridge

JOTSR:
# Summary

Encapse private methods and attributes inside a class to securise access with private fields. Support JSON handling. Fix logic and types errors. Simplify code...

# Motivation

One of the challenges of this project is the JS bridge, which is difficult to maintain for contributors and hard to discover for new users. A possible improvement is to rewrite the bridge using typescript, which has several benefits. First, it improves discoverability, as typescript can help with code completion and documentation. Second, it improves interoperability by providing type hints and safeguards. Third, as typescript is used in source code it don't affect the final execution (only JS is embedded). Moreover, for internal use, typescript can prevent errors (many references and typos in the original JS bridge have been caught by typescript) and improve project structure by enforcing types and modules.

# Main changes

- Add JSDoc to public API
- Remove duplicated code and definitions
- Remove deprecated code and browser API
- Polyfill unstable and no generalized navigation API used to handle navigation
- Fix types issue and declarations and prevent future bugs
- Add new feature to rebind webui listeners when user inject new html
- Rename public API for better clarity
- Replace dangerous eval
- Add missing declaration
- Replace deprecated use of "var"
- Update code to common modern js usage


AlbertShown:
This is really a big improvement to the project. Thank you @JOTSR 👍
We should add instruction in the readme to explain that to build WebUI we need to install `esbuild` /`xxd`

JOTSR:
Ok  I will add it later.
I'm currently dealing on resolving issue #98 by adding add set/get_title, set/get_size, set_position (get_position is a future work as no common browser API allow this).

petabyt:
> Ok I will add it later. I'm currently dealing on resolving issue #98 by adding add set/get_title, set/get_size, set_position (get_position is a future work as no common browser API allow this).

https://github.com/webui-dev/webui/pull/133 should make that easier

AlbertShown:
> adding add set/get_title, set/get_size, set_position

I don't think we should add `title` to WebUI because the end-user will use HTML `<title></title>` in all scenarios.

The APIs `get_title`, `get_position` and `get_size` can be added as TypeScript functions in the WebUI-Bridge, then we make a new C core function to call those JavaScript functions. But I vote not to do it. The end-user can write his custom JavaScript function if needed. But feel free to add them if you like too 👍

JOTSR:
As `title`, `position` and `size` is a part of the "external" aspect of the UI, proposing an turnkey solution is less boilerplate to implement for the end user. API to manage the rendered UI are out of scope obviously. 

https://github.com/webui-dev/webui/assets/58234376/9df066fa-c13b-4429-830a-9e629bdb9eaa

Issue #134, opened 2023-07-13
Creating a Release in a GitHub Action


AlbertShown:
We solved the compression `zip`/`gz` script (https://github.com/webui-dev/webui/issues/120). And we agreed that we should remove the pre-built binaries from the repo.

Why not create a new workflow task that automatically runs when a merge is tagged with `release`, so this workflow will build, compress, generate and add a new release?

hassandraga:
This is important feature to add. I already removed all the compiled objects from the repo.

ttytm:
Good to see you back @hassandraga. I hope it's going well.

Yep, this one will have a leverage. Also, for all wrapper libraries. I started working in at #155 as part of it and it's nearly finished.
We merged #152 a bit prematurely - though it does no harm, since no releases are being made yet - but once the Makefile is out of the way we can update the steps in the CI and it's done.

hassandraga:
Sounds good @ttytm

ttytm:
To share a thing that might be relevant and that I didn't came to mention in #152.

The PR intentionally didn't added the version number for the file names.
E.g. if you look at the file names:
https://github.com/webui-dev/webui/releases/tag/2.3.0 (latest release)
https://github.com/ttytm/webui/releases/tag/v0.0.1-test (WIP release on fork)

Not adding an additional version number to the release files is common e.g.:
https://github.com/neovim/neovim/releases/tag/stable
https://github.com/kovidgoyal/kitty/releases/tag/v0.29.2

This makes it easier to deal with in scripts.

E.g. it would allow to use always use an url like `/releases/latest/download/windows-msvc-x64.zip` to download the latest release without the need for traditional queries or to update it for to version changes in the file name:
https://github.com/ttytm/webui/releases/latest/download/windows-msvc-x64.zip
To download a specified version number a tag would be specified in the url
https://github.com/ttytm/webui/releases/download/v0.0.1-test/windows-msvc-x64.zip

hassandraga:
It's fine not to add the version number in the file names. We can simply add later a short description readme file in `dist` to show the current version number.

ttytm:
It's the same for the architecture in the file names. 

Adding it to downloadable archives, but keeping the filename simple.

This would make:
`macos-clang-x64.tar.gz`
`macos-clang-arm64.tar.gz`
With the file name for the static library release:
`libwebui-2-static.a`
instead of `libwebui-2-static-x64.a`

Currently, keeping just the architecture in the filename
[...]

PR #143, opened 2023-07-21
ci: rework build workflow, deploy artifacts

tobealive:
This PR reworks the build workflow to fix and extend it.

The main feature this results in is a CI of deployed build artifacts.
Example: https://github.com/tobealive/webui/actions/runs/5626014645

While working on this feature, the current build workflow revealed itself as a bit foolish. Up until now there was no testing of windows with MSVC. This can either be seen by listing the directory contents or by uploading the build artifacts after the nmake command was run - both would result in an empty output. So this issue had to be fixed within the frame of this PR. All in all I tried to add the best practices and performance optimizations that I knew of.

The changes are the major part of #134. I will build upon them in the coming days to use the produced build artifacts in a workflow for version releases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,60 +1,127 @@
-name: Build WebUI
+name: Build
 on:
   push:
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
+  setup:
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


AlbertShown:
> "Nightly" pre-compiled builds can be found as release artifacts in the Actions section

Do you mean creating releases daily and adding them to the release section?

tobealive:
Afaik nightly usually refers to the latest development build.

The workflow would add the build outputs of every commit as articats to the related workflow. So theoretically a user could always download the latest prebuild for all plafroms from the workflow.

The part I would work on separately is uploading them to github releases 
But this would only happen when a tag / version release is done.

AlbertShown:
This is amazing work @tobealive! Thank you for that 👍

Issue #114, opened 2023-05-27
Build system usability and security concerns


GiuseppeCesarano:
The current build system is difficult to use and maintain, it is too decentralized and makes it difficult to apply changes.

Running the `./linux_build` script applies many changes to the repository and it is difficult to figure out which ones should be tracked properly; by the way, with the current build system and the fact that library binaries are included in the repository, it is particularly easy for an attacker to push a build with added malware.

I strongly recommend removing previously compiled versions from git tracking because it is unusual and not a safe practice.

As for the build system itself, it should be rewritten to use fewer files and remove duplication, i.e. all clang and gcc files are completely interchangeable except for how the compiler is called.

Redundancy creates problems if, for example, a single flag needs to be changed or added. I experimented with LTO and found that enabling it didn't increase compilation time by much, and the binaries produced were 15% smaller; but setting the flag in each and every file is painful.

Finally, this system is also slow because the files are compiled sequentially.

The solution proposed in #94 is good, but I agree that zig is not that widely used yet, then I don't think there is a good integration for Visual Studio.

A single makefile would be nice but it would not be possible to support windows, I think the only solution would be to use cmake as it is a de facto standard and visual studio has good integration.

neroist:
I agree!

hassandraga:
Understandable. Honesty, WebUI has changed a lot since the first version [v1.0](https://github.com/webui-dev/webui/releases/tag/1.0.20) in November 2020.

**ASIO to C**
WebUI was completely written in C++ and used CMake as the build system. Later we find that people are struggling to install [Boost ASIO](https://www.boost.org/) libraries to be able to build WebUI v1.0. So, we decided to rewrite WebUI entirely in C.

**CMake to Makefile**
Later, I thought it would be fun to build WebUI easily as `git clone` and then `make`. That's it. Instead of installing CMake, explain to the user how to add CMake arguments to select their preferred compiler and other settings. But I'm open to bright back CMake or any other build system as an additional option.

**Adding pre-built binaries**
For a while, the Deno and Python wrappers have existed in the same main repo, so we find that Python and Deno's developers are struggling to build WebUI from C to test it... was painful, so we added the pre-built binaries to let those developers be able to test WebUI easily. Now, as we moved those wrappers to separate repos, I guess we should remove all the pre-built binaries, except the `.zip` as it is used in the website and GitHub releases.

GiuseppeCesarano:
[...]

PR #144, opened 2023-07-21
build: remove outputs from repository source files, update build scripts

tobealive:
This PR removes the build outputs and release files from the repositories source files. It keeps the `Makefile`s.

Like mentioned in some other places e.g. #114 it's not the preferred way of distributing build and release files.

From #143: "Nightly" pre-compiled builds can be found as release artifacts in the Actions section of the webui repository https://github.com/webui-dev/webui/actions. Similar like they are found on this PR branch action:
https://github.com/webui-dev/webui/actions/runs/5626124981?pr=143


AlbertShown:
LGTM

AlbertShown:
> "Nightly" pre-compiled builds can be found as release artifacts in the Actions section

Do you mean creating releases daily?

AlbertShown:
Let's discuss this [here](https://github.com/webui-dev/webui/pull/143).
About this PR, I guess it's good to be merged as we already agreed to remove all pre-built binaries.

tobealive:
> Let's discuss this [here](https://github.com/webui-dev/webui/pull/143). About this PR, I guess it's good to be merged as we already agreed to remove all pre-built binaries.

Nice, I'm at my phone atm. I'll see later if I can do it else tomorrow if there can be done something with the gitignore file so users get less changes in version control when they run the Makefiles themselves 🙂.

.gitignore
```diff
@@ -23,6 +38,7 @@
 .idea/
 *.recipe
 *.idb
+*.iobj
```
tobealive, line 41: This adds some common ignore patterns. For us it concerns the build outputs, for example when a build script or Makefile is run. It will stop files like `build/Linux/Clang/libwebui-2-static-x64.so` being tracked as changes by a users git version control when the mentioned files are executed.

The added patterns are common ignore patterns and to keep a little better overview I'm grouping them with the present patterns into categories like it is done e.g. in the [gitignore template](https://github.com/github/gitignore/blob/main/C.gitignore) for C files.

One thing I want to mention is that we still have static / shared library files in the text-editor example. They won't be ignored, since they are already part of repository and will not be deleted with this PR. Maybe we should also create a Makefile for the text-editor example, like it is done for the other examples so we can also remove them? But this would be out of the scope of this PR and should be done separately I think.

build/linux_build.sh
```diff
@@ -16,7 +16,9 @@ echo "WebUI v$WEBUI_VERSION Build Script"
 echo "Platform: Linux x64"
 echo "Compiler: GCC and Clang"
 
-RootPath="$PWD/../"
+RootPath="$PWD/.."
+BuildPath="$RootPath/build/Linux"
+DistPath="$RootPath/dist/Linux"
```
tobealive, line 21: This creates variables for the build and dist path. Both are used repetitively.

build/linux_build.sh
```diff
@@ -38,17 +38,15 @@ echo "Building WebUI using GCC...";
 echo "";
 
 # Build WebUI Library using GCC
-cd "$RootPath"
-cd "build/Linux/GCC"
+cd "$BuildPath/GCC"
```
tobealive, line 41: The variables allow for direct access instead of using:

```diff
- cd "$RootPath"
- cd "build/Linux/GCC"
```

build/linux_build.sh
```diff
@@ -59,32 +57,40 @@ cd "$RootPath"
...
-    # Release Linux Include
-    cp -f "include/webui.h" "Release/Linux/include/webui.h"
-    cp -f "include/webui.hpp" "Release/Linux/include/webui.hpp"
+    # Remove Linux distributable files directory if it exits
+    [ -d "$DistPath" ] && rm -r "$DistPath"
+
+    # Create Linux output directories
+    mkdir -p "$DistPath/include"
+    mkdir "$DistPath/GCC"
+    mkdir "$DistPath/Clang"
```
tobealive, line 74: The release directories were a fixed part of the repository and their presence did not need to be checked. After having them removed: Here, we make sure the destinations the files are copied to exist and are clean.

tobealive (commented):
Next to removing the build outputs the `.gitignore` file and build scripts were updated. Below I'll add some descriptions to the changes.

Now, when the build scripts are executed, the compiled files would be put into the `dist/` directory - the directory name is commonly used for compiled output files ([`/dist` "distributable", the compiled code/library](https://stackoverflow.com/a/22844164/18392410)). This would be created instead of the current `Release` directory - which was the only directory in the repository root starting with an uppercase letter. Please let me know if changes are desired.

.gitignore
```diff
@@ -6,9 +6,24 @@
...
+*.dll
+*.so
+*.so.*
+*.dylib
+
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
```
tobealive, line 26: This adds some common ignore patterns. For us it concerns the build outputs, for example when a build script or Makefile is run. It will stop files like `build/Linux/Clang/libwebui-2-static-x64.so` being tracked as changes by a users git version control when the mentioned files are executed.

The added patterns are common ignore patterns and to keep a little better overview I'm grouping them with the present patterns into categories like it is done e.g. in the [gitignore template](https://github.com/github/gitignore/blob/main/C.gitignore) for C files.

One thing I want to mention is that we still have static / shared library files in the text-editor example. They won't be ignored, since they are already part of repository and will not be deleted with this PR. Maybe we should also create a Makefile for the text-editor example, like it is done for the other examples so we can also remove them? But this would be out of the scope of this PR and should be done separately I think.

tobealive:
What I never asked: Do we even need to keep the build scripts when the releases are made available via GitHub? As they run the `Makefile`s and compress the outputs in a separate directory, I imagine this mainly helped the maintainers creating the releases. Maybe now it's enough to simplify it and remove the scripts. If there is a need to self-compile, users could still build the versions that fits their needs directly using the related `Makefile`.

.gitignore
AlbertShown: I vote to remove the pre-built binaries from examples too. Users can easily download the ready-to-use binaries from the Release section in Github.

tobealive:
Thanks for your reply @AlbertShown.

I see that it's good to keep them. The updates that are made to them here can then serve a good purpose as well. I would merge this then, if I may.

AlbertShown:
LGTM.

PR #148, opened 2023-07-24
docs: use github hosted assets for images in `README.md`

tobealive:
This PR will move toward using Github assets for the image files used in the readme.
This reduces the size of the repository when cloning and cleans up the source files. E.g. in the branch from which the changes are submitted: https://github.com/tobealive/webui/tree/docs/cleanup

* The text-editor example was the only file that was re-created before uploading(just a window screenshot on mac). Now it uses a transparent background to better support githubs different color themes. 
  > 

  > 

  I center aligned it, since mac adds quite some space around the screenshot to blur out their shadows, which would look imho more displaced when it's not centered.

* We could also center align the logo and badges, it looks quite nice. What do you I think?
  > Screenshot from 2023-07-22 03-51-09

<br>

**Note before merging**: This PR removes the logo files from the repository. Since it is only one of the logo files that is used here, and only in the readme, the idea is to create a `webui-dev/logo` repository for the logo files. This is similar to what was done recently with the website and also a common practice for a project logo, e.g. https://github.com/vlang/v-logo. The necessary permissions should be on my account so I can take care of creating it when the changes are approved before they are merged.

The benefit of reducing the readme asset files could also be extended to all webui wrapper repositories.


AlbertShown:
This is interesting. Yes, that's looking good to me 👍
In addition, we could remove those files from the commit history. If someone knows how without breaking the entire history commits.

tobealive:
Yes, I've done such rebases a few times. But usually on private repos before they went public as it's not recommended to rebase a public main branch. But for a few special cases like removing accidentally added data or in our case it can be the right solution. E.g. this is a repo I remember where it was done for same reasons we encounter now: https://github.com/mizlan/iswap.nvim/issues/58

I could prepare it, test the circumstances users would encounter and maybe we could push it with the next version release and a note about it?

AlbertShown:
Yes, if you have already done this, then go ahead 👍
But the rebase should not remove all commit history, just the binary files commits.

About this actual PR it seems okay to merge. Thank you.

tobealive:
Please let me know if I have overseen anything: https://github.com/webui-dev/logo

Also, if centering of the images in the README here is okay. There is no uniform style of the images we use, so it's not too easy to settle for me. But maybe this centered style is good.
You can open the current readme and the one of the fork next to each other and compare them.

AlbertShown:
[webui-dev/logo](https://github.com/webui-dev/logo) seems good!
Just fix the typo `Official Logo for WebUI |` to `Official Logo for WebUI`

> You can open the current readme and the one of the fork next to each other and compare them.
I already did, seems okay in Chrome Windows 👍

tobealive:
Just seeing, it's not possible for me to make the logo repo public. You might be able to do it, else I'll just re-create the repo as public:
Screenshot from 2023-07-24 21-03-01

AlbertShown:
I have the same access rights as you, @hassandraga is the only Admin of this project. I guess it's easier if you created a new public repo.

Issue #138, opened 2023-07-17
Keeping webui.js


AlbertShown:
We moved to TypeScript, which gave us many new possibilities. But many users that want to build WebUI wish if they can build it in a few steps as possible.

For those who do not have the ts-to-js converter tool, I suggest creating a new folder, `generated` in the `src` folder, to keep our generated `webui.js` there. So the TS/JS developers can rerun it to re-generate `webui.js`, while others can easily skip this step as the repo will have an already generated one.

**Suggestion:**
`src/client/generated/webui.js`

_Or simply `generate` or `js`._

JOTSR:
I'm not sure that tracking build output is a good solution as it could introduce confusion and mistakes into the build process. If builds are included in tracking, we can't guarantee that they strictly follow the sources. For example, if you push a tagged version without rebuilding the bridge, you could push an outdated API. Or, if someone edits only the JS and pushes it, it could add divergence between the sources and the dists, which could lead to code loss.

One solution, as suggested in #134, is to release JS within the archive.

For the same reason, we should ignore binary tracking in the repo and leave it only for the release archive.

AlbertShown:
We have two main regular scenarios:

**1:** Someone wants to download WebUI and embed it into his project. In this case, he can download it from the release section.

**2:** Someone wants to build WebUI from the source. In this case, he needs to install a `C` compiler, `npm`, `xxd`, and `esbuild`.

Right now, WebUI v2.0.3 people need only one simple C99 compiler, making the experience straightforward. I'm trying to make the coming version simple too.

I'm a C++/Go developer, and my manager allowed me to contribute to the WebUI during working hours as we are using WebUI in our products. I'm updating the core anyways to serve our needs, so at the end, I simply drop our contribution into this repo. And, no one here has `npm`, `xxd`, or `esbuild` installed.

AlbertShown:
> could introduce confusion and mistakes into the build process

I'm not sure. The build process can simply run `esbuild` to generate `output\webui.js`. If esbuild is not installed, then this step is simply skipped and the C-Build process is going to take the old existing `webui.js`.

Don't forget that only people who have `esbuild` installed can update `webui.js`. It's kind of you to update `webui.js` for us, and we just simply using it :-)

JOTSR:
[...]

PR #152, opened 2023-08-04
ci: add release step

ttytm:
<em>PR description will be updated</em>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,10 +60,13 @@ jobs:
           rm build/${{ runner.os }}/README.md
           rm build/${{ runner.os }}/${{ matrix.compiler }}/Makefile
           cp -r include build/${{ runner.os }}/${{ matrix.compiler }}
+          artifact=${{ runner.os }}-${{ matrix.compiler }}-x86_64
+          # Add the ARTIFACT name(`,,` ^= lowercase shell param) as GitHub environment variable.
+          echo "ARTIFACT=${artifact,,}" >> $GITHUB_ENV
       - name: Upload Artifacts
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


ttytm:
My time is fairly limited atm :/. I don't think to get the chance to finish it today. But over the weekend it should be done.
I'll test cross compilation for mac m1 and add it as mac release if all works out. When it's finished I'll leave a short description about the release process and mark it as ready for review :+1:.

JOTSR:
LGTM.
Maybe for a next PR but in #138 @AlbertShown discussed to track the generated js files in the repository and I proposed to rather providing them (*.js and xxd outputs) in the release source code archive. What do you think about that and is it possible to simply add these bundle in `build.yml`.

ttytm:
@JOTSR 
Yes it's possible. We build the client in the workflow so we could just add the `webui.js` to the bundle. Lets agree on something in the related PR.

ttytm:
I came to work late on this on despite the weekend so it will take a few days more. Sorry about the delay.

Something positive I can say is that building for m1 macs works. Also I think we can handle the examples nicely and make them generally cleaner. I'll push necessary changes that can stand on their own in separate PRs. So we get step by step to finishing the release CI.

AlbertShown:
@JOTSR You can close the review as done if you think this should be merged.

About the `webui.js` file, there are many solutions. Still, the main goal is to make non-js developers can easily clone and build WebUI without the need to install `dxx`, `esbuild`, `npm` etc... those should be installed and used for TS/JS developers who want to improve/debug the WebUI-JS-Bridge Script.

ttytm:
Probably a little misconception but the PR was marked as draft because it wasn't finished.

AlbertShown:
Opsy!

PR #155, opened 2023-08-09
build: use a single Makefile

ttytm:
This PR is currently WIP. 

Finished it...
- is a major step towards finishing #152.
- simplifies the build process by using a single `Makefile` in the projects root that works dynamically based on the platform and also allows to specify a compiler via e.g. `COMPILER=gcc`.
- allows to specify an `ARCH_TARGET` for mac, which I can use during github CI runs to add `ARCH_TARGET=arm64-apple-darwin` to compile for m1 and add it to the releases
- allows to remove all the `build/` directory with all its files

Either here - as it is related - or in a separate PR: I will remove the need to hold static libraries in the example directories and rather update their Makefiles to handle the static files.
- will remove the need for


AlbertShown:
@hassandraga Please invite @ttytm. Thanks.

ttytm:
Thanks for the work on the repo @AlbertShown! I tor an invitation a few weeks ago and can merge things. But I feel a bit safer submitting such changes as PR so more then just two eyes can check the changes to prevent potential oversights. I would merge things myself when it has the approval of one other maintainer or small obvious things without review. 

The PR is not finished, please don't merge yet.

README.md
```diff
@@ -80,29 +80,71 @@ Think of WebUI like a WebView controller, but instead of embedding the WebView c
 
 ## Documentation
 
- - [Online Documentation - C](https://webui.me/docs/#/c_api)
- - [Online Documentation - C++](https://webui.me/docs/#/cpp_api)
+- [Online Documentation - C](https://webui.me/docs/#/c_api)
+- [Online Documentation - C++](https://webui.me/docs/#/cpp_api)
```
ttytm, line 84: Removing the preceding white space makes the formatting uniform with the `## Features` section in the readme.

My formatter was complaining about this, I hope it's okay to include it here.

README.md
```diff
@@ -80,29 +80,71 @@ Think of WebUI like a WebView controller, but instead of embedding the WebView c
...
- - [Linux](https://github.com/webui-dev/webui/tree/main/build/Linux)
- - [macOS](https://github.com/webui-dev/webui/tree/main/build/macOS)
-
+Clone the repository
+```
+git clone https://github.com/webui-dev/webui.git
+```
+
+Build the library
+
```
ttytm, line 94: The `## Build` section is updated to conform with the upcoming changes to the building process, and also includes an info about building the client. While writing this comment I realize that the client dependencies should be added as well. I will update the last commit to include them.

To see the GitHub rendered version of the Readme please check:
https://github.com/ttytm/webui/tree/build-single-makefile-in-root

Any critique is welcome.

ttytm:
It's babysteps on the PR due to the time constraints but we'll get there soon 👍

README.md
ttytm: Another way to add build instructions would be to enable the wiki for the repo and move the build section from the Readme to the wiki and refer to it with a link. E.g., like neovim does: https://github.com/neovim/neovim#install-from-source

hassandraga:
Thank you @ttytm for this, yes, one `makefile` will be very appreciated. And This is the requirement:

### WebUI Build

- One makefile can deal with all compilers, MSVC, GCC, Clang, TCC
- Should successfully build the WebUI from source `webui/src`, and put the output in `webui/build`
- Should not deal with WebUI-Bridge at all to keep the possibility of compiling WebUI by using a simple C99 compiler

### WebUI-Bridge Build

- Should be completely separated from _WebUI Build_
- Should be able to successfully build WebUI Bridge in `webui/bridge` and put the output in the same folder

hassandraga:
I suggest to modify/new PR as this:

- Keep the one Makefile in the root
- Update Readme to match the readme in `webui/bridge`
- Move `build.sh` and `build.ps1` into `webui/bridge`
- If the one Makefile works, then we can celebrate and remove the entire folder `webui/build`

ttytm:
Thanks for throwing out all the work in the lasy days @hassandraga, it's motivating. The goal is to finish this PR here today as well.

hassandraga:
If you have any file conflicts because of the last commits, just do a fresh `clone` in a new folder, then add your makefile in `webui/makefile`. Then add `webui/bridge/build.sh`, and `webui/bridge/build.ps1`. Then push.

ttytm:
Thanks, yes I'll rebase onto the lastest main state and squash related commits to bring them in a comprehensible structure before marking it as ready. My visit just left, I'm on this now 🤓👍.

ttytm:
This might be an engineers disease to underestimate the work ever so often - especially when it comes to cross platform compatibility with different compilers. Or maybe it's just me.

My goal for today was not fully met, but I got very close.

- Linux and Macos works. 
- For windows GCC and TCC works. I will give MSVC a polish after getting some sleep.
- I removed the build section update in the Readme for now as it was outdated for the latest changes and requirements. 
- I removed the bridge build scripts from this PR and submit them separately (#164) as this is a change that can stand on it's own already.
- The examples were updated to work with the new build output paths. Rebuilding the library was removed for now. Once this Makefile - for the library itself - is finished it will be easier to update the examples to also use a single Makefile based on the finished file. Then rebuilding the library can be re-added to the examples Makefile.

To complete this PR I'm aiming to finish MSVC compilation and the Readme tomorrow, I think this is realistic this time. Other changes like a single Makefile for the examples can still be done separately, they can work with multiple until then.

hassandraga:
Thank you @ttytm for supporting Linux, macOS, and Windows.
https://github.com/webui-dev/webui/pull/164 is good to merge.
Yes, After this one is done, we'll be able to deal with examples later.

hassandraga:
- About the examples, I guess we need to remove all OS folder, so `examples/C/call_c_from_js/Linux/Clang/Makefile`, will become `examples/C/call_c_from_js/Makefile` as an example.
- About the binaries output, I suggest to completely remove the `build` folder from the project, and only use `dist` folder. So example's makefile will use `../../../../../dist` instead of `build`.

If you want, you can focus on the root makefile only, because it's the most important, when it's done, I will copy it and I will update all examples to use it.

ttytm:
Hey @hassandraga,

The build/ directory is completely removed in this PR.
Since it was already part of the repository you would need to `git rm -r --cached build/` to have it removed from your local git cache. New clones won't have it tracked.

Why not use `build/` for the output it's most common I think.

I didn't updated all until the most recent push. They should work now. Currently, the output files will be put in `build/` when building the library. After building the lib running any makefile for the examples should work.

MSVC now works aswell. `nmake` has so many complaints, best thing I could think of for nmake was adding a `Makefile.nmake`. To build with MSVC it's `nmake -f ./Makefile.nmake`.

I just get CI and the Readme finished. Please confirm what you prefer for the output dir `build` or `dist`

hassandraga:
I'm fine with keeping `build`, but then we should remove `dist`. We should not have two folders both having compiled binaries.

hassandraga:
We should also add more sub-folders like `x86-64`, `x86-32`, `ARM64`, `ARM32`. I will do it after merging this PR.

ttytm:
What is the use case for the subfolders if I may ask?

hassandraga:
To build for deferent platforms, so we can solve issues like this https://github.com/webui-dev/deno-webui/issues/9.
I'm planning to up the makefile later after we merge this PR.

ttytm:
Yes thought that it is to support differnt platforms, but I don't see yet why different directories are needed for this. 

The output of the new makefile would always land in build or dist. The architecture would depend on the system it was build on. I added the possibility to cross-compile on mac for mac m1 arm. So the output can be used in the CI and for releases.

Would adding different directories be the same as it is for the compilers now with `GCC` `CLANG` `MSVC` `TCC` directories? One intention with the single makefile was to remove the need for this.

.github/workflows/build.yml
```diff
@@ -48,26 +48,24 @@ jobs:
         uses: ilammy/msvc-dev-cmd@v1
       - name: Build
         run: |
-          cd build\/${{ runner.os }}\${{ matrix.compiler }}
+          mkdir build
```
ttytm, line 51: Windows CI made problems here when not creating the directory. The makefiles work on a local machine when the directory is not present.
Please make sure that this also works for you then I would say to `mkdir build` here as a workaround.

hassandraga:
Yes, the readme section will stay as you [suggested](https://github.com/ttytm/webui/tree/build-readme-alt#build).
but we should add the target platform in the lib file name, for example `webui-2-x64.dll`, `webui-2-i386.dll`, `webui-2-arm64.dll`, `webui-2-arm32.dll`... the subfolders in the `dist` it's just to make things more organized.

.github/workflows/build.yml
hassandraga: You can keep it for now, later we will move it to `dist`, or move all output to `build`. Whatever will save us time.

hassandraga (approved):

hassandraga (approved):

hassandraga:
I will create a Wiki page later to explain everything. This PR right now is looking good to me.

hassandraga (approved):

ttytm:
Thanks for the review! Readme was updated to the suggestion and out dirs to dist.
I'll merge it when it is green.

One thing that is left todo I see is that the msvc lib is not added to the artifacts:
https://github.com/webui-dev/webui/actions/runs/5929013692?pr=155

hassandraga:
First of all, Thank you very much for finishing this new build system, I appreciate it @ttytm 👍

> One thing that is left todo I see is that the msvc lib is not added to the artifacts

Yes, `windows-msvc-x64` has only include, and `windows-gcc-x64` has non-needed `*.o` files. We will fix this separately.

AlbertShown:
Thank you @ttytm, finaly one makefile is here 🐱‍👤 
but I can't build on Windows using GCC. Any reason why you removed `-DMUST_IMPLEMENT_CLOCK_GETTIME` ?
I fix it after this that flag to `CIVETWEB_DEFINE_FLAGS`.

ttytm:
I kept it for a while but removed it after testing it from several angles. I must need to keep this additional flag for Windows GCC. #167  Would re-add it for Windows GCC. Sorry about the inconvenience.

PR #156, opened 2023-08-10
allow faster rendering

7flash:
After compiling lib with these two flags removed, I have noticed significant improvement in performance of my app.

I understand those two flags are opposite to each other, first one disables GPU rendering and second one disables CPU rendering, my understanding might be wrong, but I don't see a reason to have both of them passed together.


AlbertShown:
The original creator of WebUI put them there... I don't know why. I guess it's better to remove them as your PR suggests. Thank you @7flash for that.

hassandraga:
Me neither I don't know why I put them there in the first place... 🤷‍♂️

PR #164, opened 2023-08-21
Add bridge build scripts

ttytm:


hassandraga:
LGTM.

ttytm:
Lemme just try to use them in CI. Then we have them covered.

ttytm:
Would work. Let me know if that okay :+1:

PR #165, opened 2023-08-21
Fix CI cache restore for Windows

ttytm:
Either it was a transitory bug with the Cachefile action or it is due to changing the shell for windows.

This PR should fix it. I make sure the change is compatible with the new Makefile in the root directory before marking it as ready.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,18 +14,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
-      - uses: pnpm/action-setup@v2
-        with:
-          version: latest
       - name: Bundle WebUI Bridge
-        shell: bash
+        if: runner.os != 'Windows'
         run: |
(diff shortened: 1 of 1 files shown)


AlbertShown:
Thanks for fixing that. I was about to fix it after I saw the new pushes get errors!

ttytm:
It looks like there is still an issue with restoring the bridge cache. But with the latest merge it's for linux.

I'll look into this further for a reliable fix.

ttytm:
From #167 
> The reason for the failing cache seems to be on the cache action side:
> 
> ```
> Warning: Failed to save: uploadChunk (start: 0, end: 13458) failed: Cache service responded with 500
> ```
> 
> Screenshot from 2023-08-21 19-55-32

I'll add a fallback for this.

ttytm:
> I'll add a fallback for this.

Reflecting on a failing third action. 
There should not be a need to implement a fallback for such things. Using GitHubs Job re-run feature is appropriate for this I think.

PR #166, opened 2023-08-21
Update table format for supported web browsers

ttytm:
- Updates formatting for the table of supported web browsers
  current: https://github.com/webui-dev/webui#supported-web-browsers
  suggested: https://github.com/ttytm/webui/tree/docs/format#supported-web-browsers
- Removes first indent level whitespace for a list to e.g. match the formatting for the `Feature section

What do you think of adding a table of contents on top of the Readme for quick navigation? Kind of like here: https://github.com/ttytm/wthrr-the-weathercrab#-wthrr-the-weathercrab

Just above the `Features` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,18 @@
 > 
 > * We are currently writing documentation.
 
+## Contents
+
+- [Features](#features)
+- [Showcase](#showcase)
+- [UI & The Web Technologies](#ui--the-web-technologies)
+- [Documentation](#documentation)
+- [Build](#build)
+- [Examples](#examples)
+- [Supported Languages](#supported-languages)
+- [Supported Web Browsers](#supported-web-browsers)
+- [License](#license)
+
 ## Features
 
 - Written in Pure C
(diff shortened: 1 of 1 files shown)


hassandraga:
I like the new table, shorter and easier to read. Thank you.

hassandraga:
Yes, why not, we can add table of contents.

ttytm:
The commit that adds the toc incorporates some related changes:

- The `Wrapper` and `Supported Langues` section were merged to reduce redundancy.
- The `UI & The Web Technologies` serves as a higher level section to related sections. With a lot of single `h2` level elements the table of contents looked like it could have been organized better. 

Please re-investigate the link above when you find a free moment and don't hesitate to me know if any of the changes should not be made or if you change your mind on the toc implementation!

ttytm:
:+1: @hassandraga 

Should we fix
`Why Use Web Browser?`
to `Why Use a Web Browser?` 
or `Why Use Web Browsers?`
or keept it as it is?

hassandraga:
Probably `Why Use Web Browsers?` is better!

ttytm:
To keep it clean, squash merge it as `Update Readme` partner commit 🙂?

ttytm:
> Sure, this will reduce the commits history. We may later need to delete build/ from history as well.

Yes if that's possible let's do it. 

But for the `build/` dir I worry that the effort might be immense, since it's history intertwined with several files and changes. It's relatively simple to keep things intact when it's something like an image or binary file.

hassandraga:
Deleting history for only binaries files, if possible, let's investigate...

PR #167, opened 2023-08-21
Keep `-DMUST_IMPLEMENT_CLOCK_GETTIME` for Windows GCC

ttytm:


Makefile
```diff
@@ -12,7 +12,7 @@ ARCH_TARGET ?=
 COMPILER ?=
 
 # Build flags
-CIVETWEB_BUILD_FLAGS = -m64 -o civetweb.o -I "$(MAKEFILE_DIR)/include/" -c "$(MAKEFILE_DIR)/src/civetweb/civetweb.c"
+CIVETWEB_BUILD_FLAGS := -m64 -o civetweb.o -I "$(MAKEFILE_DIR)/include/" -c "$(MAKEFILE_DIR)/src/civetweb/civetweb.c"
```
ttytm, line 15: no need to have this as recursive assignment

ttytm:
The reason for the failing cache seems to be on the cache action side:
```
Warning: Failed to save: uploadChunk (start: 0, end: 13458) failed: Cache service responded with 500
```
Screenshot from 2023-08-21 19-55-32

AlbertShown:
I confirm that `DMUST_IMPLEMENT_CLOCK_GETTIME` is needed only by GCC on Windows. And it's not required by GCC on macOS, Linux, or any other compilers. Only GCC in Windows.

ttytm:
Thanks for catching this so quickly and sorry about the removal. It should have been kept. Can you specify what it is needed for? Then I would take a note on it and keep it in mind for any future changes here and potentially in other projects.

ttytm:
It's probably be a version specific issue. Since the windows 10 system in a relatively virgin state and the GitHub CI didn't reported on a build fail. Whats the output of your `gcc --version`?

AlbertShown:
It's not OS-related, but the MinGW.
`gcc (x86_64-win32-seh-rev2, Built by MinGW-W64 project) 12.2.0`

If adding the flag does not break the build on your system, then it's safe to add it. The linker (LD) will ignore all non-used references.

ttytm:
Yes it's definitively right to add it, just to make reduce the change of breaking things like this again it's good to know :+1: 

Thanks for the infos @AlbertShown !

PR #168, opened 2023-08-21
Optimize CI

ttytm:
This is a workflow strategy I developed and started to incorporate into several projects recently.

It splits the platform workflows into their dedicated files. Through using callable workflows and triggering them from one master CI file it keeps all jobs organized in one Workflow and visualizes it nicely. 

This is what it would result in:
https://github.com/ttytm/webui/actions/runs/5930971332

The change comes with a little performance benefit as the setup jobs are split and consecutive jobs don't need to wait for all platforms to be setup. Another little benefit from this is that when  changes are added that make a setup for one platform fail it would continue for other platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+name: CI
+on:
+  push:
+    paths-ignore: ['**/*.md']
+  pull_request:
+    branches: [main]
+    paths-ignore: ['**/*.md']
+
+jobs:
+  windows:
+    uses: ./.github/workflows/windows.yml
+    permissions:
+      contents: write
+
+  macos:
+    uses: ./.github/workflows/macos.yml
+    permissions:
+      contents: write
+
+  linux:
+    uses: ./.github/workflows/linux.yml
+    permissions:
+      contents: write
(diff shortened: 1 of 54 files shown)


AlbertShown:
Look very interesting 👍

ttytm:
I'll also add a fix for the release step here. It's necessary to prepare an archived file for the release upload since we use a directory. Make file conditions also behave differently on mac so I simplified the condition and use what works.

hassandraga:
This is a very interesting feature to add to this repo, many people open issues just asking where to download safe pre-built binaries for their platform. I'm looking forward to finishing this PR. Many thanks @ttytm 👍

**linux-clang-x64:** OK
**linux-gcc-x64:** OK
**macos-clang-arm:** OK
**macos-clang-x64:** Missing `.dylib` file
**windows-gcc-x64:** Has `nul` and `.o` files
**windows-msvc-x64:** OK
**webui-0.0.1-test:** OK

ttytm:
Checking on https://github.com/webui-dev/webui/actions/runs/5951659214?pr=168
`.dylib` is included

ttytm:
Outputs ref: https://github.com/webui-dev/webui/actions/runs/5978637716?pr=168

hassandraga:
Seems all archives are correct now 👌
Thank you @ttytm

PR #170, opened 2023-08-22
Cleanup quotes

ttytm:
This suggestion is an attempt to make the readme even cleaner for a polished first impression.

- The block quotes on top of the readme are currently repetitive. So the first point is merged into a single initial statement.
- We are currently writing documentation is moved into the documentation section as GitHub Note highlight.

To save you a few sec navigating to the rendered preview:
current: https://github.com/webui-dev/webui
suggestion: https://github.com/ttytm/webui/tree/docs/update-note

Let me know what you think.


AlbertShown:
Personally, I prefer this [new one](https://github.com/ttytm/webui/tree/docs/update-note). It's more organized, the only thing we should add is the Pascal wrapper to the table.

hassandraga:
LGTM.

PR #174, opened 2023-08-23
Fix file deletion in Windows GCC CI build

ttytm:
Fixes the file cleanup for Windows GCC CI builds and remaining file residues in the archives.

The shell switches to bash, despite calling the make command from the default powershell. When we unignore all errors we then get `/usr/bin/bash: line 1: del: command not found`.

Includes a mini revert of a mistake in the last commit. The path should not be enquoted on declaration as we still add more segments to it in other places. Sorry about that.


hassandraga:
LGTM

PR #177, opened 2023-08-25
Use build output filename that is architecture independent

ttytm:
Step towards #134 and #168


ttytm:
Will handle this directly in #168. It's easier to add it there then first here and rebase afterwards.

PR #183, opened 2023-08-26
Prevent bridge build permission error

ttytm:
Testing the release workflow. I've run into the possibility to encounter permission errors when running the build script for the client. This change makes sure permissions are set.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -31,7 +31,7 @@ jobs:
           key: ${{ runner.os }}-${{ github.sha }}-bridge
           fail-on-cache-miss: true
       - uses: microsoft/setup-msbuild@v1.1
-      - if: ${{ matrix.compiler == 'MSVC' }}
+      - if: matrix.compiler == 'MSVC'
         uses: ilammy/msvc-dev-cmd@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/windows.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           key: ${{ runner.os }}-${{ github.sha }}-bridge
           fail-on-cache-miss: true
       - uses: microsoft/setup-msbuild@v1.1
-      - if: ${{ matrix.compiler == 'MSVC' }}
```
ttytm, line 34: Dosn't relate to permisssons. Enclosing strings `${{` and `}}` are simply not needed here.

.github/workflows/windows.yml
```diff
@@ -31,7 +31,7 @@ jobs:
...
+      - if: matrix.compiler == 'MSVC'
```
ttytm, line 34: Dosn't relate to permisssons. Enclosing strings ${{ and }} are simply not needed here.

hassandraga: Sure

hassandraga (approved):

PR #187, opened 2023-08-31
Include debug builds in CI

ttytm:
Adds building of the debug version to the CI workflow.

That way users could always download pre-built static for the latest commit from the github actions artifacts. (Related https://github.com/webui-dev/pascal-webui/issues/3#issuecomment-1699829791)

Also, if I recall it right, after moving to the single Makefile approach, you had to fix an issue with the debug build that might have been covered when `make debug` would be part of CI.

E.g.: https://github.com/ttytm/webui/actions/runs/6040804140

When creating a tag release, still only the release version would be uploaded to github releases.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -17,7 +17,7 @@ jobs:
           path: bridge/webui_bridge.h
           key: ${{ runner.os }}-${{ github.sha }}-bridge
 
-  build-release:
+  build:
     needs: setup
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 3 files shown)


hassandraga:
This is one of the most important CI honestly, this will help users testing WebUI easier, and faster. Thank you @ttytm very much for this.

hassandraga (approved):

hassandraga:
Can we rename `macos-clang` to `webui-macos-clang` ?

ttytm:
Ofc I missed that one. Good catch @hassandraga 👍 . Updating it when I'm back at my desk, at the phone atm.

hassandraga (approved):

PR #197, opened 2023-09-05
Use a single Makefile to build examples

ttytm:
Moves towards a single Makefile (+ a Makefile.nmake for Win MSVC) for examples.

Implemented / Tested:

- [x] GCC Linux
- [x] Clang Linux
- [x] Clang macOS
- [x] GCC Windows
- ~~TCC Windows~~
- [x] MSVC Windows

Other TODOs:

- [x] Readme update
- [x] C++ Makefile update

Updating the C++ Makefiles will be a simple job based on the C Makefiles, just finishing C first.


hassandraga:
That's good. To make things simpler and easier I suggest removing the `.ico` resources support.

ttytm:
Okay good :+1: 

What I want to add: This PR is based on the assumptions, that it's not part of the repo that the Makefiles in the examples serve as examples of how to write a Makefile for a platform. The content of the single Makefiles becomes more complicated here. But the changes should make the examples easier accessible and simpler to compile with different compilers. I hope this is more beneficial to the repo.

hassandraga:
I suggest to completely remove the TCC support as it need headers to work properly.

ttytm:
@hassandraga I'm torn between using a the same `Makefile` per example directory, to just run the make command e.g. `make`, OR a `Makefile` in the examples root directory (`examples/C/` / `examples/C++`) and run the make command + <example_name> as target, e.g., `make call_c_from_js`. Do you have a preference? 

`make` from the directory of a single example
`make <example_name>` from the examples root directory

hassandraga:
I prefer `make` from the directory of a single example as it's simpler and easier. I already saw your PR and I was about to try it first, but now as you are asking, I really prefer a simple makefile for each example. Removing OS folders from examples is already a huge improvement.

ttytm:
Okay thanks @hassandraga. 

It's basically done, I'll just do some more test later or tomorrow and add a thing about the examples that I could observe during working on this that might need to be addressed before marking this as ready.

ttytm:
Other changes added:

- `CC` (e.g. `make CC=clang`) to specify a compiler, as it's more common and shorter
- `BUILD_LIB` (i.e. `make BUILD_LIB=true`) can be use to build / re-build the WebUI library when compiling an example.
- Added a Makefile to fix building of the text-editor, added build support for linux and macOS.
- ~~The final executable is portable and has less than _1 MB_ in size (_+html and css files_).~~ Removed from the text-editor readme as it seemed obsolete. The executable is not portable without the UI files (on windows it also needs the .dll file next to it) and the bins size is ~45kb, the html,css files are ~3MB.

hassandraga (approved):

ttytm:
Just added an update of the screenshort as part of the last commit.

hassandraga:
LGTM. Ready to marge. And thank you for this PR @ttytm 👍

FYI:

> The executable is not portable without the UI files

I guess _portable_ means copying the example folder (_exe + html_) to another machine and running it is possible. You don't need to install any runtime or any dependency.

> on windows it also needs the .dll file next to it

That's because the dynamic build version, the static one won't need any external lib.

hassandraga:
The examples are cleaner now. I like it. Thank you agin @ttytm

ttytm:
Thanks :pray: please don't hesitate to update and change where I missed things. E.g. 
I've seen I left the line about size and portability in the examples/C/README.md.

> I guess portable means copying the example folder (exe + html) to another machine and running it is possible. You don't need to install any runtime or any dependency.

I see, it indeed is portable, just took the "executable" too literally as the single executable file.

PR #199, opened 2023-09-05
Update images with wide widths to fill content

ttytm:
This includes a minor cosmetic change.

It updates the "wide" images in the readme to re-created versions. The main goal was to make all wide images fill a readmes max content (1012px is max at GH afaik). 

For the other changes, I hoped the updated cppcon snapshot is in initially a bit simpler and more interesting and that the Diagram image transmits the point of how easy it is to create an application with full OS API access (it's obvious when thinking about it, now it's in the image as well). The Diagram also won't use a shadow anymore to keep things uniform.

Quick link to rendered preview: https://github.com/ttytm/webui/tree/readme/update-images

The current state is also good, and those are things where preferences might differ, so please tell if the changes are a +.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ This [text editor example](https://github.com/webui-dev/webui/tree/main/examples
 
 <div align="center">
 
-![CPPCon](https://github.com/webui-dev/webui/assets/34311583/cf796ead-66d3-4298-ac80-b551c25f3e41)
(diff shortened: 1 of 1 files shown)


hassandraga:
I like more your updates over the existing one 👍

hassandraga (approved):

PR #200, opened 2023-09-05
Remove TCC related code

ttytm:
Makes the `Makefile` a lot cleaner as different compilers only need to be handled for Linux now.


README.md
```diff
@@ -102,9 +102,6 @@ Think of WebUI like a WebView controller, but instead of embedding the WebView c
   # GCC
   mingw32-make
   
-  # TCC
-  mingw32-make COMPILER=tcc
-
```
ttytm, line 107: What is left in the readme is:
> - Lightweight (~160 Kb using TCC) & Small memory footprint

How should this be updated? Didn't knew at first glance what things are calculated.

hassandraga: We can keep it as the user can still add the headers files and compile the lib using TCC to get the 160Kb lib. But let's change it to `~200Kb` as GCC/Clang produces 200Kb files.

The main goal here is to explain to the user that WebUI is a KBS lib and not MBs lib compared to other GUI libraries.

hassandraga (approved):

README.md
ttytm: So you say?
Lightweight (~200 Kb using GCC/Clang) & Small memory footprint

Or
Lightweight (~200 Kb using TCC) & Small memory footprint

hassandraga: Lightweight _~200 Kb_ & Small memory footprint

PR #201, opened 2023-09-08
Improve debug build target usage

ttytm:
Using a distinct file for debug builds would simplify some things imho. There are a bunch of cases where it makes sense to have both files side by side.

If you also see value in this, I'd finish it here.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,7 +25,6 @@ jobs:
     strategy:
       matrix:
         compiler: [GCC, Clang]
-        kind: [release, debug]
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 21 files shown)


hassandraga:
I already taught about adding the debug version to the CI build output. Good that you bright this on.
Is `dist/webui-2.dll` and `dist/debug/webui-2.dll` can solve the issue?

hassandraga:
I mean, yes, it's good to continue, but I suggest keeping the file name the same, but simply add a `debug` folder inside the `dist` folder. `dist/debug`. If it's not possible then we can add `-debug` to the output file name.

ttytm:
I see. That's good. And it should work!

ttytm:
What do you think about shipping both in then in the releases and artifacts instead of separating them there

```
webui-linux-gcc-x64.zip
├── debug
│   ├── libwebui-2-static.a
│   └── webui-2.so
├── include
│   ├── webui.h
│   └── webui.hpp
├── libwebui-2-static.a
└── webui-2.so
```
something like this?

Maybe this is event what you mean by saying
> I already taught about adding the debug version to the CI build output. 

because we actually already have Debug builds in the CI output.

hassandraga:
> What do you think about shipping both in then in the releases and artifacts

Yes, that what I meant 👍
one archive for both release and debug

ttytm:
Testing the changes during the integration already feels like this change makes a lot of things simpler. So I feel reinforced that this is a good step to take.

The commits could have been split more fine grained, as they contain some other small fixes. But I hope until we have arrived at a new stable version some crammed up commits are okay, mainly due to time reasons.

hassandraga (approved):

PR #203, opened 2023-09-12
Implement nightly releases

ttytm:
Implement nightly releases that are automatically updated when changes to the main branch are made.

Screenshot_20230912_045520

This change should work very well as it is, but there are always improvements to make.

E.g. possible future improvements:
- Making releases could be split into it's own `release` workflow and file. This would require some re-organisation, since artifacts would need to be downloaded to make them accessible for a separate job and the release preparation would need to be done there, after downloading.
- What triggers a nightly release could be narrowed down further to changes that are decided to be relevant (the separate workflow mentioned above would be supplementary for this).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,11 +68,33 @@ jobs:
         with:
           name: ${{ env.ARTIFACT }}
           path: ${{ env.ARTIFACT }}
-      - name: Prepare Release Archive
-        if: github.ref_type == 'tag'
-        run: tar -czvf $artifact.tar.gz ${{ env.ARTIFACT }}
-      - name: Release
-        if: github.ref_type == 'tag'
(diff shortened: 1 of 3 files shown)


hassandraga (approved):

hassandraga:
Other improvements to add in the future:

- Windows ARM64 (_Needed for Qualcomm & Microsoft Surface devices_)
- Linux ARM64 (_Needed for Raspberry Pi devices_)

ttytm:
We can probably update the link for the latest beta build to https://github.com/webui-dev/webui/releases/tag/nightly

hassandraga:
We thought about it at the same moment!
https://github.com/webui-dev/webui/commit/c50a576b906e2e83b083635b233d5ac6f9cb2668

PR #217, opened 2023-09-22
Add `ChromiumBased` to `webui_browsers` enum

ttytm:
Chromium based browser share the same api.
So I'm running into the use case and that might be common to just use a chromium based browser, no matter which one specifically.

This PR would add a `ChromiumBased` enum field to use a any chromium based browser.


hassandraga (approved):

hassandraga:
Thank you for this PR, Yes, `` is an usefull case 👍
I did a quick update:
- Moving `ChromiumBased` to `user specified browser` section
- Update the order of `ChromiumBased`

Now seems good to me!

PR #219, opened 2023-09-22
Fix disabling translate feature on Chromium based browsers

fibodevy:
"--disable-translate" doesnt work since few years. For Chrome to offer a translation just refresh the WebUI with Ctrl+R (maybe several times). The text in the browser must be in a language different than yours of course.

"--disable-features=Translate" fixes it, but I would keep "--disable-translate" too.

wB8f4Yszoy


ttytm (approved):

hassandraga:
Good catch! 👍
LGTM

PR #226, opened 2023-09-23
Improve text editor example

ttytm:
Two improvements are made. Currently, saving a file is only enabled when a file is opened. This is accompanied by that it is not possible to save text that is simply typed into the editor - without a file having been opened before.

The PR:
- Enables the save button when a change is made in the editor. It will also be disabled again after saving a file until the next change is made.
- Allows saving text entered in the editor when no file has been opened and sets the file data (doctitle and language) after saving.
- Makes some minor refactors.

### DIFF
--- a/examples/C/text-editor/ui/index.html
+++ b/examples/C/text-editor/ui/index.html
@@ -1,33 +1,23 @@
 <!doctype html>
 <html>
     <head>
-        <meta charset="UTF-8">
+        <meta charset="UTF-8" />
         <script src="/webui.js"></script>
 
         <link rel="icon" type="image/png" href="img/icon.png" />
         <title>Text Editor in C using WebUI</title>
-        <link rel="stylesheet" href="css/style.css">
-        <link rel="stylesheet" href="css/lucario.css">
-        <link rel="stylesheet" href="css/codemirror.min.css">
-        <link rel="stylesheet" href="css/all.min.css">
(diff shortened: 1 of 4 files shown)


examples/C/text-editor/ui/index.html
```diff
@@ -1,33 +1,23 @@
...
-                <li>
-                    <a id="close-button"><i class="fas fa-circle-xmark"></i></a>
-                </li>
-                <li>
-                    <a id="about-button"><i class="fas fa-question-circle"></i></a>
-                </li>
-            </ul>
+            <button id="open-btn" title="Open file" type="button"><i class="fas fa-folder-open"></i></button>
+            <button id="save-btn" title="Save file" type="button" disabled><i class="fa fa-floppy-disk"></i></button>
+            <button id="__close-btn" title="Close file" type="button"><i class="fas fa-circle-xmark"></i></button>
```
ttytm, line 19: I'm opting for the double underscore prefix syntax for ids used by the backed. I think it looks better combined with regular conventions than mixing casing and is more explicit and better recognised imho.

E.g.
```html
<button id="save-btn" ...
<button id="__close-btn" ...
```
vs
```html
<button id="save-btn" ...
<button id="CloseBtn" ...
```
At fist I thought about a `webui__` prefix but liked `__` best as it should be explicit enough.

fibodevy (approved):

hassandraga:
LGTM

PR #230, opened 2023-09-24
sep24 - second

7flash:


fibodevy (changes requested):
What'cha doin? This is a UI library, what goal are you trying to achieve? You mentioned you need to "talk" with an extension, may I ask what for?

Anyway this isnt how it should be done. If you have such specific requirements you should use your fork.

7flash:
> What'cha doin? This is a UI library, what goal are you trying to achieve? You mentioned you need to "talk" with an extension, may I ask what for?
> 
> Anyway this isnt how it should be done. If you have such specific requirements you should use your fork.

Thanks for review. It's specifically useful for web3 apps which depend on user wallet installed as extension. Moved PR into separate fork.

PR #241, opened 2023-10-07
Fix a typo at `webui.c:4511`

clzls:
Fix a typo at `webui.c:4511`
`browser` here is ridiculous since we may assert `browser == AnyBrowser` in this if-branch.


ttytm (approved):
LGTM

fibodevy (approved):

hassandraga:
Good catch 👍

PR #244, opened 2023-10-08
Add formatter config and workflows, format project

ttytm:
Adds changes mentioned in: https://github.com/webui-dev/webui/discussions/242

I think the settings resemble the current style fairly well.

A style I couldn't find a matching setting for is to indent everything in pre-processor directives. Adding `IndentPPDirectives: BeforeHash` to `.clang-format` would come closest, but variable declarations still wouldn't be indent, resulting in:

- Uniform indent without decls:
  ```c
  #if defined(__clang__) || defined(__GNUC__)
  	#define WEBUI_DISABLE_OPTIMIZATION_START _Pragma("GCC optimize (\"O0\")")
  	#define WEBUI_DISABLE_OPTIMIZATION_END
  #elif defined(_MSC_VER)
  	#define WEBUI_DISABLE_OPTIMIZATION_START __pragma(optimize("", off))
  	#define WEBUI_DISABLE_OPTIMIZATION_END   __pragma(optimize("", on))
  #else
  	#define WEBUI_DISABLE_OPTIMIZATION_START
  	#define WEBUI_DISABLE_OPTIMIZATION_END
  #endif
  ```
- Not uniform indent for decls:
  ```c
  #ifdef _WIN32
  static const char* webui_sep = "\\";
  static DWORD WINAPI _webui_run_browser_task(LPVOID _arg);
  static int _webui_system_win32(_webui_window_t* win, char* cmd, bool show);
  static int _webui_system_win32_out(const char* cmd, char** output, bool show);
  static bool _webui_socket_test_listen_win32(size_t port_num);
  static bool _webui_get_windows_reg_value(HKEY key, LPCWSTR reg, LPCWSTR value_name, char value[WEBUI_MAX_PATH]);
  
  	#define WEBUI_THREAD_SERVER_START DWORD WINAPI _webui_server_start(LPVOID arg)
  	#define WEBUI_THREAD_RECEIVE      DWORD WINAPI _webui_receive_thread(LPVOID _arg)
  	#define WEBUI_THREAD_RETURN       return 0;
  #else
  ```
  
- So I left the setting out to use the default / more commonly used style of not indenting pre-processor directives. The used sections for the directives are usually highlighted by the editor if an lsp is running:
  
  Screenshot_20231008_032820

If any changes would further improve things for you let me know. 

Example runs:
- https://github.com/ttytm/webui/actions/runs/6444738382 [success]
- https://github.com/ttytm/webui/actions/runs/6444423773 [fail]


ttytm:
As Microsoft is in the space nazi camp when there is no config like this, here is where you can configure tab width on github to use what you like best for reviews on the page: https://github.com/settings/appearance 

> Screenshot_20231008_035944

ttytm:
Currently the lint assumed that clangd installed used locally, then all common editors could take care of formatting via format on save or running a format command.

I didn't integrated formatting and writing to the repository if formatting is off. We can do it if that's preferred.

hassandraga:
Formating on CI is better as we can not guarantee that all contributors have Clang installed.

ttytm:
Okay, I see that automated writing of formatting changes can make sense in our case. I'll look into it.

Though it is be a valid requirement for contribution to meet a projects code style. And it's not guaranteed for most formatters unless a formatter is built into the language itself. Alternatively, a contributor could format manually to satisfy linting for now.

hassandraga:
I mean, whatever is easier and efficient we should do it. We can format PR manually.

hassandraga:
I will use Clang to format code for my next PRs.

PR #245, opened 2023-10-10
Use GNUmakefile and Makefile to simplify MSVC compilation

ttytm:
Next to just running `make` and `mingw32-make` as usual for compiling lib and examples, the change allows to also just run `nmake` instead of `nmake -f Makefile.nmake` to compile with MSVC.


hassandraga (approved):

hassandraga:
I like this one 👍
Thank you

PR #251, opened 2023-10-11
`WEBUI_MAX_ARG` fix

fibodevy:


hassandraga:
Thank you for fixing the bug. That's actually fixing an overflow crash 👍

All lines are correct, except this one should back to original as Token is not related to call arguments, and it has already a bigger size.

`while (token != NULL && token_num < WEBUI_MAX_ARG+1)`

hassandraga:
For next PRs we should start formatting code using Clang, I will start doing it too.

PR #252, opened 2023-10-11
Add pre-commit config

ttytm:
Adds a pre-commit git hook for clang format.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v16.0.6
+    hooks:
+      - id: clang-format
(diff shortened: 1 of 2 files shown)


hassandraga (approved):

hassandraga:
That's mean `CI / lint / clang-format (push)` will be removed from PR check list?

ttytm:
Nope, it'll still be there.

This just adds a config for the pre-commit git hook. It's something that will be run locally if a contributor has the pre-commit tool installed when a git commit is made. Then clang format will be run over the committed changes and format them if necessary. So it's just a save guard in case format on save wasn't turned on, or the code wasn't formatted for any other reasons when something was committed.

For the lint workflow I think it's good to see the lint CI failing when code isn't formatted, means it's working :D

But as common as it is to have such a linting and a formatting tool running, I feel it might cause some inconveniences at the time. So the fully automated approach of automatically formatting via CI in case something was committed and not formatted, might really be the way to go. Gonna submit something next days :+1:

hassandraga:
> the lint CI failing when code isn't formatted, means it's working :D

That's the deal! 🤷‍♂️

> automatically formatting via CI in case something was committed and not formatted

The main maintainers may use Clang formatting for the next PRs, which I will do. But other contributors may not have any formatting tool so the automatic formatting via CI will be the best option if we can implement it.

PR #255, opened 2023-10-13
Create mac archive before uploading

ttytm:
Reverts the recent PR, that had no effect on the outcome.

Creates the compressed archive before uploading as an attempt to solve an issue for macos-x64 where the uploaded release archive seems to get corrupted. In case the upload action has an effect on the later archived files this should prevent it.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -50,6 +50,8 @@ jobs:
           cp -r include dist
           # Add the ARTIFACT name(lowercased) as GitHub environment variable.
           artifact=webui-$(echo ${{ runner.os }}-${{ matrix.compiler }}-${{ matrix.arch }} | tr '[:upper:]' '[:lower:]')
+          chmod -x dist/webui-2.dylib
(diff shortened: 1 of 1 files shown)


ttytm:
Those are pretty blind shots, don't like to approach it that way. I rather try to get my hands on an old mac check if the issue is reproducible and create the releases on a repo fork and push things when appear to be a real fix for the issue.

For now the `.dylib` for macOS x64 nightly releases seems to be broken.

hassandraga:
While I'm re-testing the security side of the `.dylib` file, we can check the `GNUSparseFile.0` issue meanwhile.

ttytm:
:+1: Okay, I'm also removing the executable bit, the reduce the chance of mac flagging it as insecure. The bit is also removed automatically when the artifact is uploaded in the actions. 

Maybe one of the changes here works. We can try it.

> we can check the GNUSparseFile.0 issue meanwhile.

Is also something I cannot reproduce. So currently I have to delegate fixing this.

hassandraga (approved):

PR #256, opened 2023-10-14
Update README.md

7flash:


README.md
```diff
@@ -155,6 +155,9 @@ Think of WebUI like a WebView controller, but instead of embedding the WebView c
 | Apple Safari    | _not available_ | _coming soon_ | _not available_ |
 | Opera           | _coming soon_   | _coming soon_ | _coming soon_   |
 
+### Who's integrating WebUI
+[Galaxy Browser](https://github.com/galaxydo/galaxy-browser) - interactive canvas-based knowledge management system, it allows users to write any javascript and python scripts directly on the canvas, then invoke them to be executed in Deno backend via WebUI, showing results as text and images appearing straight on the canvas.
+
```
ttytm, line 160:
```suggestion
### Who's integrating WebUI

[Galaxy Browser](https://github.com/galaxydo/galaxy-browser) - interactive canvas-based knowledge management system, it allows users to write any javascript and python scripts directly on the canvas, then invoke them to be executed in Deno backend via WebUI, showing results as text and images appearing straight on the canvas.

```

7flash: > doc format linting would fail without the linebreak. Need to extend prettier to .md only changes.

thanks fixed

hassandraga:
Thank you for the link. I already taught about that, and I was planning to add this section after we release `2.4.0` and wait until three or more significant projects use it so we can add their links as examples, including galaxy-browser. But not now in the beta 👍

PR #279, opened 2023-12-07
optimize workflows - remove unused builds

7flash:


ttytm:
Hey @7flash

The builds are used in two ways

1. To test building with the specified compiler
2. To give the option of using the build output that was build with another compiler

ttytm:
Thanks for the willingness to improve things. As it is useful to keep the builds I'm closing the PR.

PR #283, opened 2023-12-30
add `build.zig` to support zig build system

jinzhongjia:
I had completed wrappping of `webui` in `zig`

Add this `build.zig` to allow zig treat c `webui` as a library

This allows us to use zig's package management system more efficiently

Note: I just complete basic building in this `build.zig`, not include examples building. If necessary, I will initiate a new PR or commit later.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,0 +1,156 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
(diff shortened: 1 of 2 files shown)


jinzhongjia:
Of course, we can use zig as C/C++ compiler in CI.

PR #297, opened 2024-01-05
webui chromium proxy support

osljw:
remove chromium_options --no-proxy-server, export webui_set_proxy api


fibodevy:
How about other browsers? API functions should work for all supported browsers, at least for all major browsers.

This is Chromium specific. To make it more general I would suggest making APIs to add new parameters and exclude existing parameteres, eg `webui_startup_param_add(window, param)` and `webui_startup_param_exclude(window, param)`, so one could add or exclude browser startup parameters. @hassandraga how about that?

PR #306, opened 2024-01-14
fix: add linux dynamic library "lib" prefix

iacore:


neroist (approved):

ttytm:
In case this change is made, wouldn't it be most consistent make a general update, instead just the dyn out on linux? 

https://github.com/webui-dev/webui/blob/9f120e974fcb333bbc749790ea233c15e5494513/GNUmakefile#L5

iacore:
> In case this change is made, wouldn't it be most consistent make a general update, instead just the dyn out on linux?
> 
> https://github.com/webui-dev/webui/blob/9f120e974fcb333bbc749790ea233c15e5494513/GNUmakefile#L5

That is not what this is about. Only the Linux .so output is broken, I think.

Also, did someone forget to merge this? :?

ttytm:
I don't think it was broken either, it was just the name chosen so far. We can look into platform naming conventions get to the best possible state making a breaking change like this.

boussou:
Hi, this PR should be accepted.

On linux ld standard is to prefix by lib.

And the during the build it warns: 

`/usr/bin/ld: note to link with /usr/local/lib/webui-2.a use -l:webui-2.a or rename it to libwebui-2.a`

PR #310, opened 2024-01-20
Draft: Add CMake build system

NicolasBFR:


NicolasBFR:
For some reason the linking works fine on all Linux compilers but fails on the `_mg_websocket_write` symbol linking. Anyone have any idea ?

jinzhongjia:
Do you mean this problem with cmake ?
I am currently developing under arch linux, and the original makefile and zig builds are fine

NicolasBFR:
> I suggest providing `CMakeLists.txt` only while keeping the original CI (_GitHub Workflow_) untouched as it's working fine now. cmake is like the zig build, is an extra if anyone needs it.

@AlbertShown 

I just wanted to make sure it would technically work for all platforms, unfortunately there is a linker error and I have no Mac or Windows installation to debug it.

CMake is obviously not necessary for a small project like this, but I needed it to import the webui target in a bigger project (AI inference application). 

It builds fine on Arch Linux and Ubuntu, should I remove the CI changes, squash history and remove the draft status of the PR ?

jinzhongjia:
Don't worry, we need to wait @hassandraga  for a decision

eliemichel:
> On the other hand, webui is simple, one .c file, so cmake may be overkill, in my opinion.

This is really not transparent from reading the README. If using WebUI is as simple as adding a single .c file in a project, then this should be advertised in the README instead of only showing an opinionated build process.

PR #338, opened 2024-03-19
Headers and debug improvements

mitchcapper:
This has 3 commits can submit them separately or remove any not wanted but some of the lines were too close to avoid conflicts otherwise.

Most of these changes are minor, adding better MSVC debug flags, switching some char* to const char* when they are const.   Adding names to the interface_bind function pointer.  In addition to making things a bit more clear it assists autogeneration of the code with a tool like CPPSharp.  More details on that below.

The final item that is here is adding a magic cookie to event structs.   I dislike the idea of expanding the struct to include a memory check but it is one of two ways I thought to do this effectively.    Events are only executed in serial one at a time and do not support callbacks.  This may make users more likely to use them in an async fashion and potentially try to set the result after they should (or call another function on them).   As we try to use the event ID directly this could cause use after free issues (but with something like interfaces may not be invalid due to the event buffer).   

I thought about just adding this with a DEBUG compile time flag, but changing the event struct size based on that seemed like a bad idea.   The only other option takes advantage of the serial call limitation.   We could store the pointer to the current event struct, and the current event's id (for interface callbacks).   We could then assert if the pointer/id number that come in don't match.   

### Auto Code Generation
 With these changes no modifications are needed to generate a fairly 1:1 api: https://github.com/mitchcapper/WebUIAutoBuild/blob/master/CodeGen/WebUIGeneratorDirect.cs

resulting in: https://github.com/mitchcapper/WebUIAutoBuild/blob/master/Generated/WebUI.cs

With a more complicated generator can generate it into a more standardized class form: https://github.com/mitchcapper/WebUIAutoBuild/blob/master/CodeGen/WebUIGeneratorClass.cs

I put together a quick sample for what each looks like when used at:
https://github.com/mitchcapper/WebUIAutoBuild/tree/master/ConsoleSampleApp

but for the class version its pretty spot on:

Note these were not to replace the actual c# lib ( https://github.com/Juff-Ma/WebUI.NET ) but I generated for some testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ TLS_LDFLAG_DYNAMIC = libssl.lib libcrypto.lib
 CIVETWEB_BUILD_FLAGS = /Fo"civetweb.obj" /c /EHsc "$(MAKEDIR)/src/civetweb/civetweb.c" /I"$(MAKEDIR)/src/civetweb/" /I"$(WEBUI_TLS_INCLUDE)"
 CIVETWEB_DEFINE_FLAGS = /D NDEBUG /D NO_CACHING /D NO_CGI /D USE_WEBSOCKET $(TLS_CFLAG)
 WEBUI_BUILD_FLAGS = /Fo"webui.obj" /c /EHsc "$(MAKEDIR)/src/webui.c" /I"$(MAKEDIR)/include" /I"$(WEBUI_TLS_INCLUDE)" $(TLS_CFLAG)
+VS_DEBUG_ARGS = /ZI /Zf /FS /JMC
 
 # Output Commands
 LIB_STATIC_OUT = /OUT:"$(WEBUI_OUT_LIB_NAME)-static.lib" "webui.obj" "civetweb.obj"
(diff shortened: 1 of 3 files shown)


mitchcapper:
> but char * url seems redundancy code, does it suppress a MSVC warning?
Officially a const char * shouldn't change, and we need to copy something into it after assignment.  If there was a _webui_strdup would certainly work.  Still I can avoid the extra code and just quietly cast the  const char * to a char* for that scoped and we won't tell anyone. We can't just use a char * for window_url in general as we assign content to it, which is a const (although again could cast it).   So of the two slightly hacky things i think I prefer the cast into strcpy.

> If the MAGIC_COOKIE purpose is to add assert(e->magic_cookie == MAGIC_COOKIE);, I guess if(event_inf == NULL)return; is already doing that,

Not quite.  A user could hold on to an event until a random time after execution and then attempt to set the result on it despite it being freed.    `assert(event_inf != NULL)` at least makes sure an event is executing so certainly better.   Still this library has the benefit of single thread synchronous event handling and we know the event we are handling.  My recco would me to ditch the magic cookie and instead we simply `assert (event_inf == e)`.    Should be pretty foolproof without adjusting the struct.  If it sounds good I will update the PR.

PR #340, opened 2024-03-20
Context menu control

mitchcapper:
This adds setContextMenuEnabled to the webui object for users to be able to turn it on and off.  I think the best reason to allow a user to use a 3rd party context menu like: https://github.com/fkhadra/react-contexify

A few other minor things:
*) Used passive event binding for situations where we don't modify the event object
*) Moved the context menu bind off body to the document itself.  Otherwise you can right click on the lower part of the window with shorter pages.
*) To avoid duplicating code I moved the load handler inside the main class.  There may be some specific reason it wasn't, but I didn't see what it was.


mochalins:
Thanks for the PR! I'll get to rebasing/reviewing this in a bit. However, I have to comment that unfortunately any public facing API changes require a review from a core maintainer; I understand it's been quite a while since this PR was made, but I'll try to flag someone down to get this reviewed and and merged once I've gotten it updated. Thanks for your patience.
@mitchcapper

Issue #335, opened 2024-03-11
`.mjs` files are responded with `text/plain` rather than `text/javascript`


sitd2813:
**On**: deno-webui 2.4.4 (webui 2.4.2)
**Setup**: `index.html` with `<script type="module" src="main.mjs"></script>` as entry point

`.mjs` is a common filename extension for a JavaScript module file along with `.js`. However, WebUI does not seem to recognize this. If I use `main.js`, it works, but `main.mjs` doesn't.

### Error message

```text
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
```

jinzhongjia:
Indeed this is a problem, maybe we can open an additional API later to specify mime

PR #343, opened 2024-03-24
Add MIME type for `.mjs`

sitd2813:
Resolves <https://github.com/webui-dev/webui/issues/335>


sitd2813:
Civetweb added support for `.mjs` here (<https://github.com/civetweb/civetweb/issues/1252>), so as long as that someone uses the lastest `civetweb.c` from upstream, I think it would be fine.

fibodevy (approved):

PR #349, opened 2024-04-21
[Update]: update to zig nightly 0.13 support

jinzhongjia:
No changes for code logic, just adapt to zig `nightly`

Now, zig `0.12` has been released, zig `nightly` is `0.13`

And I need some test for this PR!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -11,6 +11,7 @@ const log = std.log.scoped(.WebUI);
 
 const build_11 = @import("build_11.zig").build_11;
 const build_12 = @import("build_12.zig").build_12;
+const build_13 = @import("build_13.zig").build_13;
 
 const min_zig_string = "0.11.0";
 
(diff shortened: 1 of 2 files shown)


jinzhongjia:
OK, I have test, I think all thing are ok!
Can you approve this PR? @neroist

ttytm:
Wheres the issue with using one build file? To me it looks similar to the zig-webui workfows lots of redundancies. Imho it could be simpler to read and maintain when putting some effort into more intelligent code design and structure

jinzhongjia:
> Wheres the issue with using one build file? To me it looks similar to the zig-webui workfows lots of redundancies. Imho it could be simpler to read and maintain when putting some effort into more intelligent code design and structure

I separated the build files to ensure compatibility and maintainability, because zig's build system api has changed a bit in the last two versions, making it easy to break the availability of the other versions when updating the logic if we put all the code in one file once nightly changes

As for the zig-webui workflows, I copied the webui exactly

jinzhongjia:
Of course, my current design does have redundancy, so maybe I should just say that they are divided into two files

jinzhongjia:
The benefit of the current scheme is that I only need to update 0.13.
Now 0.13 is same as 0.12, but I'm not sure whether zig's build system api will change
After all, it hasn't released 1.0 yet

jinzhongjia:
> Wheres the issue with using one build file? To me it looks similar to the zig-webui workfows lots of redundancies. Imho it could be simpler to read and maintain when putting some effort into more intelligent code design and structure

Very good, I have now found a way to solve the above problem. We only need to keep a build.zig later to ensure 0.11 compatibility.
zig adds the ability to use C packages non-invasively in 0.12

PR #353, opened 2024-05-03
[Zig Change]: remove other build.zig

jinzhongjia:
This PR is to solve the problem: https://github.com/webui-dev/webui/pull/349#issuecomment-2067969209

Now we only need to keep a `build.zig`, which exists to ensure compatibility with `0.11`

Versions `0.12` and later can directly reference the C library and compile it from scratch


jinzhongjia:
@neroist Can you approve this?

This may be the last PR about zig I initiate to this repository

fibodevy (approved):

Issue #321, opened 2024-01-28
Ubuntu 20.04 is not supported?


Zizaco:
First of all, thanks for the great work. webui is an innovative project that brings a fresh approach for making UI with web technologies.
Now, the issue that I have is: **Ubuntu 20.04 has glibc 2.31**.

When I try to use webui I get the following error:

```
Error: Could not open library: Could not open library: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/zizaco/Projects/deno/deno-webui/src/webui-linux-gcc-x64/webui-2.so)
```

I guess Ubuntu 20.04 is simply not supported then? From v2.4.2 - v2.0.2, all of the versions required glibc `2.34` or `2.33`.
Unfortunately that goes on the contrary of the promise of being "Multi-platform & Multi-Browser" and "Fully Independent" of webui.

Is there any chance of a release that work with glibc 2.31?

jinzhongjia:
you can use zig to compile this, it should use glibc 2.31 correctly

just download zig `0.11` or `nightly`, then git clone `webui`, goto the directory, and run `zig build -Dis_static=false -Denable_tls=true`, this will compile webui wtch tls support(dynamic link library)
The build results can be found under zig-out/lib

jinzhongjia:
according to this issue: https://github.com/ziglang/zig/issues/4459, zig should support glibc 2.31

jinzhongjia:
I think it's a better choice to use zig as a toolchain build, but that's up to @hassandraga and @ttytm decide
However, if webui itself does not provide libraries compiled by zig, it may be possible to provide libraries compiled by zig in the downstream zig-webui.

Zizaco:
indeed, zig is [well-known for allowing cross-compilation of popular projects](https://www.youtube.com/watch?v=5_oqWE9otaE). I found it a very reasonable approach. @hassandraga @ttytm I would love to hear your thoughts on this

halildgn:
> I think Zig is an excellent build system, but webui is very simple to build. It has one `.c` file. You can build webui by simply cloning and running `make`. So, asking all developers to install Zig is an overkill.
> 
> On the other hand, some developers want to integrate webui as a submodule in their repo, so some need Zig, others need Cmake, Ninja... that's why it is a good thing to have those build systems as well.
> 
[...]

PR #357, opened 2024-05-08
change ubuntu version in linux workflow to support legacy glib

halildgn:
As discussed in [#321](https://github.com/webui-dev/webui/issues/321), the people who run WebUI on Ubuntu 20.04 or lower get the error of version `GLIBC_2.33' not found`  since they have `glibc 2.31`(for the case of Ubuntu 20.04) or lower. Currently, a workaround has to be used as explained [here](https://github.com/webui-dev/webui/issues/321#issuecomment-2075075790). To solve this problem, Ubuntu version in the linux workflow should be changed to `ubuntu-18.04` as proposed [here](https://github.com/webui-dev/webui/issues/321#issuecomment-1914961370).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   setup:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
       - name: Bundle WebUI Bridge


halildgn:
> Thank you for the fix 👍

Thank you for everything again @AlbertShown

neroist:
Apparently the Ubuntu 18.04 runner is no longer supported... 

https://github.com/actions/runner-images/issues/6002

halildgn:
> Apparently the Ubuntu 18.04 runner is no longer supported...
> 
> [actions/runner-images#6002](https://github.com/actions/runner-images/issues/6002)
> 
> Why not try 20.04 instead, since that seems to be the version that the issue arises on.

Thank you for the guidance. It was mentioned in the comments of the issue that newer versions of Ubuntu always supports also the glib versions of legacy Ubuntu versions, so I wanted Ubuntu-18.04 users to also benefit the library if there are any. Since it is not possible, I adjusted the change for `ubuntu-20.04` as you suggested.

neroist (approved):

neroist:
LGTM

halildgn:
> Can anyone confirm that Ubuntu `latest` is using `GLIBC_2.31`?

I can check on a 24.04 VM if my [repo](https://gitlab.com/halildgn/automating-slr) that uses the version that was compiled on Ubuntu 20.04 works on Ubuntu 24.04, would it be sufficient? I checked last time with 22.04 , it was working.

ttytm:
> [...] so I wanted Ubuntu-18.04 users to also benefit the library if there are any.

Ubuntu 18.04 reached end of life for over a year soon. So it's fine to not bother about supporting it anymore. 20.04 was already superseded by two lts releases so is fine to cover "old" systems.

halildgn:
> To double-check, someone should download Linux version [nightly build](https://github.com/webui-dev/webui/releases/tag/nightly), then runs one of those commands:
> 
> ```
> readelf -V ./webui-2.a | grep 'GLIBC_'
> readelf -a ./webui-2.a | grep 'GLIBC_'
> strings ./webui-2.a | grep GLIBC_
> objdump -p ./webui-2.a | grep 'GLIBC_'
> ```
> 
> same thing with `webui-2.so`.

To make sure, you mean that I need to run one of these commands for both  `./webui-2.a` and `webui-2.so`  on Ubuntu 24.04, right?

halildgn:
> According to [#358 (reply in thread)](https://github.com/webui-dev/webui/discussions/358#discussioncomment-9458499)
> 
> The nightly build is using GitHub Ubuntu 20.4, wish is using GLIBC_2.33. So, to use webui with old GLIBC versions, the user should re-build webui from source.

20.04 uses 2.31 though

PR #361, opened 2024-05-24
Add QuickJS Wrapper

xland:
Add QuickJS Wrapper


xland:
Thank you for your support.

I initially considered naming the project `QuickJS-WebUI`.
But this name is too long.
And I intend to provide additional support to this project.
Such as accessing the system clipboard, creating tray icon, reading and writing files, etc.
Which means it won't just be a wrapper.
So I choose `QuickUI` as the project name.

ttytm:
I'll come back to this PR soon and also make a suggestion about adding this and more related projects.

Since a recent system update, I'm encountering network errors when I try to use Webui on my main system, which I couldn't isolate this morning.

PR #363, opened 2024-05-26
Don't return silently when zig build is not possible

ttytm:
Informs about the reason for an unsuccessful build when using the latest zig version.
The exit code 1 makes the build better verifiable in regular and automated scenarios.


jinzhongjia (approved):
You're right, the building result shouldn't return silently

PR #374, opened 2024-05-30
Update variable type in _webui_round_to_memory_block to resolve sing-comp warning

ttytm:
The PR enables sing-compare errors. And will attempt to solve some of them when there is time.

We should eventually resolve all the warnings/errors to follow best practices and to make things more robust.

Some of them already cause problems when trying to use webui source files directly with some C interop languages.

I think lots are easy fixes. If someone wants to append to this PR, we can do it as a tandem work. Or, if it is preferred to take one category of the errors and submit them as separate PR it will also be helpful.


ttytm:
To improve the approach on this, I'm going to close #374 and open an issue instead of a PR. There I'll also try to add better instructions on how people can help and test.

Doing it that way should be less blocking and allow us to progress step by step.

PR #376, opened 2024-05-30
Update variable type decl in internal function `_webui_round_to_memory_block` to resolve sing-comp warning

ttytm:


ttytm:
Nice you fixes already most of the warnings! Good work :100:

PR #378, opened 2024-05-30
Prevent warnings from reminder definitions defined in civetweb thridparty

ttytm:
This prevents warning from definitions in `civetweb.c` that are meant as reminders.

Those definitions can also trigger for the uses malloc

https://github.com/webui-dev/webui/blob/ce3dce4991919941431624292693821e5f1dcfdc/src/civetweb/civetweb.c#L1534-L1535

E.g., they prevent to directly interact with the WebUI C code from CGO. 
On the left an example CI run on a Go fork. On the right using webui with the civetweb.c changes in this PR can sucessfully compile:

| | |
|-|-|
|Screenshot_20240530_191931|Screenshot_20240530_192627|

The fix is achieved by wrapping the reminders in a `NDEBUG` condition which is a definition we already add during compiling.


ttytm:
Rebased over master to extend the tests on the fork. 

The PR or a similar fix seems still necessary. Using e.g. `#undef DO_NOT_USE_THIS_FUNCTION__USE_mg_realloc` for all the civetweb definitions unfortunately doesn't work, so the condition might be a slim way to solve it.

ttytm:
I see your point, and would also prefer not to change them. But sometimes it's inevitable to make minor adaptions in third-party code when working with it. Then it just can be documented what changes need to be made when the libraries are updated. 

E.g. https://github.com/facebook/zstd is used as a thirdparty module in V. When it gets updated the definition below needs to get added manually. Thirdpary updates don't happen to often, so as long there are no drastic changes it's little effort to manage.
```c
#if defined(__TINYC__) && defined(\_WIN32)
#undef ZSTD_MULTITHREAD
#define ZSTD_NO_INTRINSICS
#endif
```

When opening this PR I also submitted a PR at civetweb simultaneously. Maybe it gets integrated over there soon, then this condition comes nativeley with civetweb.

ttytm:
Nope, still when trying to interact with WebUI directly. 

```
❯ go run ./examples/minimal.go
# github.com/webui-dev/go-webui/v2
In file included from ./cgo.go:16:
./webui/src/webui.c: In function ‘_webui_is_firefox_ini_profile_exist’:
./webui/src/civetweb/civetweb.c:1558:18: error: implicit declaration of function ‘DO_NOT_USE_THIS_FUNCTION__USE_mg_snprintf’ [-Wimplicit-function-declaration]
 1558 | #define snprintf DO_NOT_USE_THIS_FUNCTION__USE_mg_snprintf
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./webui/src/webui.c:1103:13: note: in expansion of macro ‘snprintf’
 1103 |             snprintf(full_path, sizeof(full_path), "%s/%s", home, & path[1]);
      |             ^~~~~~~~
./webui/src/webui.c: In function ‘_webui_free_mem’:
./webui/src/civetweb/civetweb.c:1557:14: error: implicit declaration of function ‘DO_NOT_USE_THIS_FUNCTION__USE_mg_free’ [-Wimplicit-function-declaration]
 1557 | #define free DO_NOT_USE_THIS_FUNCTION__USE_mg_free
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./webui/src/webui.c:3124:13: note: in expansion of macro ‘free’
 3124 |             free(ptr);
      |             ^~~~
./webui/src/webui.c: In function ‘_webui_malloc’:
./webui/src/civetweb/civetweb.c:1554:16: error: implicit declaration of function ‘DO_NOT_USE_THIS_FUNCTION__USE_mg_malloc’ [-Wimplicit-function-declaration]
 1554 | #define malloc DO_NOT_USE_THIS_FUNCTION__USE_mg_malloc
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./webui/src/webui.c:3216:17: note: in expansion of macro ‘malloc’
 3216 |         block = malloc(size);
      |                 ^~~~~~
In file included from ./cgo.go:18:
./webui/src/webui.c:3216:15: error: assignment to ‘void *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
 3216 |         block = malloc(size);
      |               ^
```

The changes might still be a viable way. Best would be if they merge it over a civetweb. There is no activity atm.

Also, would it be good to add a condition to surpess the warnings to keep the possibility to get more information from the compiler during development?

PR #387, opened 2024-06-02
Use singular form for enum names

ttytm:
The PR is mainly about making signatures more expressive.

- ~~Updating the return type of `webui_get_best_browser` to `webui_browser` instead of size_t.~~
- Updating enum names to a singular form. Since an enum represents _one_ constant with the same name, it is a closer description of the value that should be passed/received. For example, `config`/`browser` instead of `configs`/`browsers`.

While the C enum names are used in the public function signatures to indicate types, there is no direct "outside" usage of the actual C enum names. So, afaik, this change shouldn't cause any breaks for one of the wrappers and should just improve expressiveness.

### DIFF
--- a/include/webui.h
+++ b/include/webui.h
@@ -578,7 +578,7 @@ WEBUI_EXPORT bool webui_set_port(size_t window, size_t port);
  *
  * @example webui_config(show_wait_connection, false);
  */
-WEBUI_EXPORT void webui_config(webui_config option, bool status);
+WEBUI_EXPORT void webui_set_config(webui_config option, bool status);
 
 // -- SSL/TLS -------------------------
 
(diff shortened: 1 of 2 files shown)


ttytm:
I see, and good points. I would paddel back on the first point and also prefer to keep it the size_t to not require API changes in the wrappers.

I'd remove the first of the commits then and just keep the name changes if that's fine.

ttytm:
At the current updated state the updates would include a break. 
Changing `webui_config` to `webui_set_config` to remove a conflict in names. 
Alternatively, using as a different enum name would work without breaking changes.

Since webui_config is a freshly introduced function it could be okay to fine tune it before the next release, it would match with other function names that `_set_`. Let me know your preference please.

neroist (approved):

Issue #388, opened 2024-06-02
Regession with the latest changes


ttytm:
Test moving back the last commits to 4694a2938931d814e0f318f5def2ef470b962f00
https://github.com/ttytm/v-webui/actions/runs/9340128030/job/25705316290

Test at current main 4466bcfc4e6ecfe92e244068d4bfbb1f5b231b59
https://github.com/ttytm/v-webui/actions/runs/9340147044/job/25705345755

ttytm:
Great work thank You!

Issue #380, opened 2024-05-31
Errors compiling minimal example: "undefined reference to `pthread_create" and "undefined reference to `dlopen'" (and my fixes)


vindarel:
Hi,

Compiling webui worked instantly, but compiling the minimal example gave me these errors:

```$ make                                                               *[main]
Build C Example (gcc release static)...
/usr/bin/ld: home/vince/lisp-projects/webui/examples/C/minimal//../../..//dist/libwebui-2-static.a(webui.o): in function `_webui_show_window':
webui.c:(.text+0x9c03): undefined reference to `pthread_create'
/usr/bin/ld: webui.c:(.text+0x9c0f): undefined reference to `pthread_detach'
/usr/bin/ld: webui.c:(.text+0x9daf): undefined reference to `pthread_create'
/usr/bin/ld: webui.c:(.text+0x9dbb): undefined reference to `pthread_detach'

[…]

/usr/bin/ld: /home/vince/lisp-projects/webui/examples/C/minimal//../../..//dist/libwebui-2-static.a(webui.o): in function `_webui_wv_show':
webui.c:(.text+0xce4e): undefined reference to `dlopen'
/usr/bin/ld: webui.c:(.text+0xcec6): undefined reference to `dlopen'
/usr/bin/ld: webui.c:(.text+0xcf1e): undefined reference to `dlsym'
```

I found a combinaison of characters in the Makefile that makes the compilation succeed and the example run:

```
# in examples/minimal/GNUMakefile
# BUILD FLAGS
# STATIC_BUILD_FLAGS = main.c -I"$(INCLUDE_DIR)" -L"$(LIB_DIR)"
STATIC_BUILD_FLAGS = main.c  -Wl,--no-as-needed -ldl  -I"$(INCLUDE_DIR)" -L"$(LIB_DIR)"
# DYN_BUILD_FLAGS = main.c -I"$(INCLUDE_DIR)" -L"$(LIB_DIR)"
DYN_BUILD_FLAGS = main.c  -Wl,--no-as-needed -ldl  -I"$(INCLUDE_DIR)" -L"$(LIB_DIR)"

[…]

else
	# STATIC_BUILD_FLAGS += -lpthread -lm -l$(WEBUI_LIB_NAME)-static
	STATIC_BUILD_FLAGS +=  -lm -l$(WEBUI_LIB_NAME)-static -lpthread
	# DYN_BUILD_FLAGS += -lpthread -lm
[...]

PR #394, opened 2024-06-04
Use semver compliant version specification

ttytm:
Hoping this PR is not too pedantic, it would update the version specification.
It might pave the way for consistent future version specifications discussions are welcome.

Goals/Reasons: 
- Follow semver syntax more closely https://semver.org
- Currently, the development state is fairly fluent. It cannot really be pinned down to a version. A `2.5.0-beta` is more general than a `2.5.0-beta.1`.  
  I suggest making pre-release after #380 and #388
  Then a version and a tag like `2.5.0-beta.1` could be used and the state pinned down to this version. Else, until a `2.5.0` release keeping it general to just `2.5.0-beta` should be okay.
  
It still looks good rendered in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ![Logo](https://raw.githubusercontent.com/webui-dev/webui-logo/main/webui_240.png)
 
-# WebUI v2.5.0-Beta-1
+# WebUI v2.5.0-beta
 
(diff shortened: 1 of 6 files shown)


neroist (approved):

ttytm:
There wasn't a clear semver yet, so I opted for paddling back to the general version at the current state. Though I don't have a strong preference here and can agree with the comment. Gonna update.

ttytm:
The update sets `beta.1` in the defined version. It removes the version string from the C comments. Having a version definition without additional comments would be less places to updated and is how e.g., civetweb or mbedtls are handling it. If there are further suggestions, let's find what suits best.

Issue #85, opened 2023-04-24
Firefox Issue


hassandraga:
If you use Linux, and your only installed web browser is the latest Firefox, WebUI probably won't work because Firefox itself can't create a new profile.

`firefox -CreateProfile "WebUI /home/me/.WebUI/WebUIFirefoxProfile"`
```
Gtk-Message: 20:15:00.605: Not loading module "atk-bridge":
The functionality is provided by GTK natively. Please try to not load it.
Error creating profile.
```

malisipi:
![image](https://user-images.githubusercontent.com/87281783/235734225-344a6e7b-d3cb-45e5-9dff-4a6817a57b07.png)
Firefox Version: `Mozilla Firefox 112.0.2` (The latest version of firefox for arch linux)

It's still working for me. It could a error that affect some versions of firefox. Or it could be existing profile bug. Idk, however i'm sure about the problem is not from we.

> But we need to add new tweaks into `user.js` to hide bookmarks

AlbertShown:
Yes, It seems to be a [Firefox issue](https://askubuntu.com/questions/1462267/when-launching-firefox-get-error-not-loading-module-atk-bridge), not WebUI.

You are right. The `user.js` and `userChrome.css` still need to be tweaked to make the Firefox window looks better.

hassandraga:
Same issue on macOS.

![macOs_WebUI_Issue](https://user-images.githubusercontent.com/16948659/236607328-34dc04ba-523d-4783-bd04-6f06fa2dfe1a.png)

hassandraga:
I'm not planning to fix this issue in v2.4.1. I hope someone have free time to deal with `user.js` and `userChrome.css` to make Firefox UI look better.

malisipi:
Webview is good idea however there are a lot of Webview library in Linux. GTK' webview, QT' webview and more. It would be difficult.

We can download Chromium browser in background by library and run application with the browser.

However i think we should support more browsers to have more options. So, I think might be reconsidered Gnome Web https://github.com/webui-dev/webui/pull/225. (Epiphany is a web browser that developed by Gnome Team.)

wwderw:
> Webview is good idea however there are a lot of Webview library in Linux. GTK' webview, QT' webview and more. It would be difficult.

Statistically speaking the GTK one is the one that I have found most to use, when it's just regular users downloading that dependency.  Qt webview is usually those that are using the Qt toolkit for the application as well.

[...]

PR #398, opened 2024-06-06
Remove platform-dependent CSS and use browser titlebar on Firefox

malisipi:
I removed platform-dependent CSS and used Firefox default titlebar support.
- Since Titlebar is handled by platform, titlebar is looking more consistent with system.
- Since removed platform-dependent CSS, should be fixed issues (like #85) and code should be more future-proof.
- Untested on Windows and MacOS yet. _If you test the PR on the platforms, can you share your test results._

resim

TODO:
- [ ] Removed browser vendor name from titlebar. _I need to run code on Browser Console **(should not be confused with Web Console)** and I can not find how to run via command-line._
- [ ] Manage title (until accessing browser console) and icon of browser with HWID of browser window, in Windows.

Tested on:
- [x] Linux
- [x] MacOS
- [x] Windows

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -4367,6 +4367,7 @@ static bool _webui_browser_create_new_profile(_webui_window_t * win, size_t brow
             );
             fputs("user_pref(\"browser.shell.checkDefaultBrowser\", false); ", file);
             fputs("user_pref(\"browser.tabs.warnOnClose\", false); ", file);
+            fputs("user_pref(\"browser.tabs.inTitlebar\", 0); ", file);
             fclose(file);
 
             // userChrome.css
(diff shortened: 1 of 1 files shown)


malisipi:
Actually if can we handle the browser window' HWID from PID in Windows, we can change title and browser logo by manually.

malisipi:
> Right now, your PR makes Firefox title-bar looks native, and it's way better than the actual userChrome.css style, We can deal with - Mozilla Firefox later...

> You can verify, and merge when you want.

So, I will merge the PR. I will search solution for titlebar, I will create a new PR if I found a good solution.

neroist (approved):

PR #401, opened 2024-06-09
fix: zig nightly becomes `0.14.0`

jinzhongjia:


jinzhongjia:
No content has been changed, just adding 14 compatibility as previously agreed.

ttytm:
What do you think on dropping support for zig `0.11` with the zig `0.14` release to focus on the newer build system? It likely gets further extended until then, and using newer features while keeping `0.11` could become a bigger limitation then.

jinzhongjia:
I think it is ok, for me, 0.11 can be said to be a semi-finished product, because the build system was only introduced in 0.11

PR #403, opened 2024-06-09
Fix conflict with including dirent

ttytm:
The PR should resolved the last changed required to satisfy the go compiler when directly using WebUI as a submodule.

It will allow for: https://github.com/webui-dev/go-webui/pull/35

On linux, dirent is always included:
https://github.com/webui-dev/webui/blob/561c0449244f60b7bfdb2b877b17242c238f12e7/include/webui.h#L75-L76

### DIFF
--- a/include/webui.h
+++ b/include/webui.h
@@ -47,9 +47,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-#if defined(__GNUC__) || defined(__TINYC__)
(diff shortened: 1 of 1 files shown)


ttytm:
I was wrong about it being the last required change. While the changes allow me to build on Windows locally, there is still a conflict when running in it in the workflows.

Next time I will do such test before making such claims. But we'll getting closer.

PR #407, opened 2024-06-12
Fix windows TLS CI

ttytm:
Uploads the TLS Windows build as release artifact. 

For now only the working GCC build.

To verify the TLS build, it can be found as an artifact in the workflow run of the PR.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -21,8 +21,6 @@ jobs:
         with:
           path: bridge/webui_bridge.h
           key: ${{ runner.os }}-${{ github.sha }}-bridge
-      - name: Install OpenSSL
-        run: choco install openssl -y
 
   build:
     needs: setup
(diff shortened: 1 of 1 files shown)


ttytm:
> I'm trying to actually include the secure version in the archives.

Yep this is what the PR does. I'm just failing to communicate:

Screenshot_20240612_175736

https://github.com/webui-dev/webui/actions/runs/9485767608?pr=407

ttytm:
It will be part of the releases as well.
When TlS building with msvc is resolved we can include it as well

ttytm:
I would give a last refinement to the PR in a sec

.github/workflows/windows.yml
```diff
@@ -31,15 +29,18 @@ jobs:
...
-          - compiler: MSVC
+          - cc: gcc
+            make: mingw32-make
+          - cc: msvc
             make: nmake
-            tls_include: "C:\\Program Files\\OpenSSL\\include"
-            tls_lib: "C:\\Program Files\\OpenSSL\\lib"
+        exclude:
+          - cc: msvc
+            secure: tls
```
ttytm, line 43: > Sure, I can't see the screenshot as GitHub servers recently has traffic issues... however, I'm working on webui\src\webview/WebView2.h(62): error C2061: syntax error: identifier 'ICoreWebView2AcceleratorKeyPressedEventArgs'.

Prepped it in a way that when the fix is finished you only would need to remove this exclude then also the msvc tls builds will be added.

Line 33 can be removed then too, it is obsolete without the `exclude`.

In case you don't see the screenshot, it shows the artifacts in the workflow summary, including the uploaded tls artifact. Also added a link to the summary to directly check it in the browser.

.github/workflows/windows.yml
```diff
@@ -21,8 +21,6 @@ jobs:
         with:
           path: bridge/webui_bridge.h
           key: ${{ runner.os }}-${{ github.sha }}-bridge
-      - name: Install OpenSSL
-        run: choco install openssl -y
```
ttytm, line 25: Removing since openssl is installed by default on runner images (ref. https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md)

Also, having this in this separate job doesn't affect the build job. The build job runs without this installation already. The effect of having this is only adding a delay to the CI

ttytm:
Good job, gonna rebase this one and enable msvc tls builds

ttytm:
> I guess it will be better if we include the `secure` version of WebUI in the same archives we already have instead of creating new archives for TLS.

Okay, what name should we choose then for the libfile?

ttytm:
I see, I think then we can supersede this PR with #409

PR #408, opened 2024-06-12
Limit releases to webui origin repo

ttytm:
This will prevent creating a release if a contributed pushes to the main branch on a fork.


ttytm:
Since the msvc tls builds are fixed this can be merged independent of the linked PR.

PR #409, opened 2024-06-12
Simplify artifact building on linux and macos

ttytm:


ttytm:
Right, they slipped in with an issue during moving things around when updating

ttytm:
Should be fixed with the latest commit

PR #413, opened 2024-06-13
Include TLS builds in artifact uploads on Linux and macOS

ttytm:


ttytm:
It would leave out TLS builds when cross-compiling to Linux arm due to openssl.

Maybe just add it when github adds its arm runners compile on arm native?

ttytm:
Or it might work to when we move to Zig for the builds. It is usually better in handling C than C 😅

ttytm:
> I don't mind using Zig to generate pre-built binaries, but many contributors know Makefile/CI scripts and not Zig.

Yes we should definitely keep and continue to support and test this build path. Also not making Zig a requirement to build WebUI.

What I meant was just for building and cross compiling the binaries in CI runs that will be uploaded as releases.

ttytm:
Zig gets a bit of a reputation of being the better C also when just using C from Zig, it appears this is somehow becoming a meme atm.. Looking into it, it's achievements are quite impressive. More companies start using Zig to do cross-compilation. I'm getting a bit of a fanboy too :D.

PR #415, opened 2024-06-15
Fix condition for release steps when pushing tags

ttytm:
I simplified the release condition too much with one of the recent updates. We would encounter an issue on the next tag release (I already did in another project after making a similar change).

While creating a tag is a push event, the ref_name is not the branch name like main but the tag's name. So, the condition would be false when it shouldn't be.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -127,7 +127,9 @@ jobs:
         with:
           tag_name: nightly
       - name: Release
-        if: github.repository_owner == 'webui-dev' && github.ref_name == 'main' && github.event_name == 'push'
+        if: >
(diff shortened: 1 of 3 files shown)


ttytm:
Yes I noticed the update. I don't have experience with github paid runners yet. There were always solutions without them. It'll be interesting to explore their functionalities. Let's see where things go.

ttytm:
Solidifying the basics and improving testing in general is needed. The main repo still has no tests besides building. I think it will yield more benefits before starting endevours on more platforms.

ttytm:
Does the runners cost per usage? I think it's better to ask first before I also start doing tests.

PR #416, opened 2024-06-15
Fix compiling with swift due to incompatible civetweb structname

ttytm:
The structname `ah` is hindering compilation with swift on macos. 
A fix can be achieved through a simple rename.

I will also make this proposal over at civetweb, as I think the longer name is a better struct name, while `ah` works well for a variable. Though idk if and how long it will take for it to get merged.

### DIFF
--- a/src/civetweb/README.md
+++ b/src/civetweb/README.md
@@ -25,3 +25,8 @@ Changes currently required when updating the submodule to ensure full direct int
   #define snprintf DO_NOT_USE_THIS_FUNCTION__USE_mg_snprintf
   + #endif
   ```
+
+- Allow compiling with swift compiler chaning an incompatible structname
+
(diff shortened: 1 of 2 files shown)


ttytm:
Errors. 
It's not possible to use swift-webui with this struct in civetweb.
In the CI of swift-webui continue-on-error is enabled for macos, checking the build step the failure is still visible. Ref.: https://github.com/webui-dev/swift-webui/actions/runs/9525013964

The fix can be verified by cloning the swift-webui repo as it's described in the readme. Trying to build it on macos and encountering the failure. Then applying this change to civetweb.c and trying to building it again.

fernandolguevara:
@hassandraga `struct ah` is colliding with [`Darwin.netinet6.ah`](https://github.com/apple/darwin-xnu/blob/main/bsd/netinet6/ah.h#L41)

ttytm:
Thanks @fernandolguevara.

The PR at civetweb is open, though the repo is inactive atm.

I'll extend the description with your provided info and ping the author, maybe we'll have and upstream fix soon then as it's a trivial one.

PR #421, opened 2024-06-16
Add tests

ttytm:
This PR adds tests to better prevent regressions.

The test can also be run locally. In the added `tests` directory.

```sh
make call_c_from_js.c
```
```
./call_c_from_js
```


ttytm:
Went fast here as it helped to continue work.

Things can be better structured eventually, but I think the tests can be a good improvement already. Don't hesitate to make updates if necessary.

PR #422, opened 2024-06-16
Format workflows, add blank lines between steps

ttytm:
The changes allows to have some flags set as env variables by only setting default variable values if they are not set. The current way of building should continue to work as is it currently does.


ttytm:
Since workflows have grown lager in the recent days. I would would do minor updates to their structure in their followup. This should improve their speed, maintainability and also make it in general more cost effective in case of using paid runners.

PR #423, opened 2024-06-16
Utilize arm runners, include TLS arm builds

ttytm:


ttytm:
It appears like they run, I would expect that the CI doesn't start when they are used as a `runs-on:` value. But I can't tell 100% since this is my first experience with those runners too. 

But there is an issue with wow the PR current arm buils, downloading an arm artifact and checking its arch it's not an arm, but a x64 binary.

I had the expectation that a native arm gcc results in an arm binary. Apparently that is not the case here. Would need to investigate further.

ttytm:
The platform info of the `buildjet` runner prints it's an x64:
```
Linux 6.5.0-1021-azure x86_64 GNU/Linux
```

Screenshot_20240617_043847

ttytm:
Maybe next to the settings of the organization they need to be enabled for the repository :thinking: Since you have full access, can you verify?

https://github.com/webui-dev/webui/settings/actions/runners

Issue #432, opened 2024-06-23
Accessibility Request: High Contrast Management


malisipi:
## Motivation

Most of browsers has a high contrast mode that enables if OS is themed by a high contrast theme. Supporting browsers default behavior is good for most usage cases however some devs might want to create custom high contrast theme to ensure high contrast theme is not breaking some UI. Because auto-theming is not working fine for all web applications. It has some problems;
* Low contrast due to unexpected re-theming. If application is using dark theme and OS has light high contrast theme, it will cause some **light** icons will placed on **light** backgrounds. (It also causes reverse, app is using light theme however OS is using dark high contrast theme)
* No borders on some popups, dialogs and UI objects. It may some confusion between users, You may not understand the dialog where starts/ends and it will not help about accessibility.
* Some applications needs different text colors than high contrast text color, for example code editors like VSCode. VSCode's high contrast theme has support for syntax highlighting.

## Description

So WebUI needs a option to detect and change behavior of browser about high contrast. It may help give more accessibility for users and more freedom for developers.

|Same App Example without High Contrast Theme|Same App Example with High Contrast Theme|
|-|-|
|![resim](https://github.com/webui-dev/webui/assets/87281783/a554b28f-a4e4-4bc9-9da8-a8d987fbb6b6)|![resim](https://github.com/webui-dev/webui/assets/87281783/3ef327d8-b4bf-4657-b5e5-4641d817aa5c)|

## Example code

I wrote some API examples for this situation. _(API names and behavior should be re-considered to save consistent.)_
```c
(webui::window)->set_browser_high_contrast_support(bool is_enabled){
    // Enables/Disables the browser support
};
bool webui_is_high_contrast_enabled(){
    return bool; // Returns OS high contrast theme enabled
    // Will return OS high theme state, ***not browser support state***
};
```

```js
function webui_is_high_contrast_enabled(){
    return bool; // Returns OS high contrast theme enabled
    // Will return OS high theme state, ***not browser support state***
}
```

malisipi:
Edge and Chrome is supporting the high contrast mode as I tested. However I am pretty sure about most of Chromium browsers and Firefox are supporting that by default. We should a flag to disable high contrast support.

To make WebUI independent, we can read GTK and KDE config files about the theme and if we detect high contrast enabled, we can report it to the application to disable or change CSS.

I would work about the API support.

malisipi:
[...]

PR #433, opened 2024-06-24
High Contrast Management

malisipi:
To get more information about why it's required, you can browse #432.

- [x] Support to enable/disable default browser behavior about high contrast.
  - [x] Chrome
  - [x] Edge
  - [x] Firefox
- [ ] Detect high contrast theme enabled
  - [ ] Windows
  - [ ] Linux
  - [ ] MacOS _(I have not a Mac computer)_
- [ ] API to check OS high contrast theme from
  - [ ] Backend
  - [ ] Frontend

### DIFF
--- a/include/webui.h
+++ b/include/webui.h
@@ -279,6 +279,16 @@ WEBUI_EXPORT bool webui_show_wv(size_t window, const char* content);
 WEBUI_EXPORT void webui_set_kiosk(size_t window, bool status);
 
 /**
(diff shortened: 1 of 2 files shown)


malisipi:
I am not much knowledge about WebUI Frontend API. So I couldn't implement the API however since backend is exist of API, it should not be hard.
Also I have not a Mac computer, so I can not implement detecting on MacOS.

Issue #430, opened 2024-06-23
Update Wrappers to 2.5.0.Beta.2


hassandraga:
- [ ] Deno
- [ ] Rust
- [ ] Go
- [ ] Swift
- [ ] Pascal
- [x] Zig
- [ ] V
- [ ] Nim
- [ ] Odin
- [ ] Python
- [ ] Bun
- [ ] PureBasic

tvortsa:
Wait Deno wrapper o.o

ttytm:
I'm pretty occupied time wise atm, sorry about delays helping out here.

Trying to use latest webui with the v-wrapper, I'm running into complications:
```
error: passing argument 1 of ‘_webui_generate_cookies’ from incompatible pointer type [-Wincompatible-pointer-types]
 7430 |             _webui_generate_cookies(new_auth_cookies, WEBUI_COOKIES_LEN);
      |                                     ^~~~~~~~~~~~~~~~
      |                                     |
      |                                     char **
/home/t/.vmodules/vwebui/src/webui/src/webui.c:584:43: note: expected ‘char *’ but argument is of type ‘char **’
  584 | static void _webui_generate_cookies(char* cookies, size_t length);
      |                                     ~~~~~~^~~~~~~
/home/t/.vmodules/vwebui/src/webui/src/webui.c:7431:49: error: passing argument 2 of ‘_webui_client_cookies_save’ from incompatible pointer type [-Wincompatible-pointer-types]
 7431 |             if (_webui_client_cookies_save(win, new_auth_cookies, &new_client_id)) {
      |                                                 ^~~~~~~~~~~~~~~~
...
```

PR #442, opened 2024-07-06
Resolve incompatible ptr warning to fix compatibility with gcc-14

ttytm:
Fixes: https://github.com/webui-dev/webui/issues/430#issuecomment-2211735415

- The error is reflected in the CI of the first commit.
- The second commit contains the quickfix.
- To prevent such issues and make them visible immediately, it would help to eventually remove the `-w` from the `WEBUI_BUILD_FLAGS` (replacing it with, e.g., `-Wall`). Imho., warnings that are within the sphere of influence of the library and not caused by third-party should be resolved instead of suppressed.

  https://github.com/webui-dev/webui/blob/836e1d115f090e1cccc2415ebb5e0b5d3b11cef9/GNUmakefile#L43

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -40,6 +40,7 @@ jobs:
       fail-fast: false
     env:
       ARTIFACT: webui-linux-${{ matrix.cc }}-${{ matrix.arch }}
+      CC: ${{ matrix.cc }}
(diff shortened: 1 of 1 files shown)


ttytm:
> Yes, why not, but only for debug version.

Good idea

PR #444, opened 2024-07-07
Changes to support c-embed

nater0000:
Added a helper function to retrieve the server port for the session webui_get_port()
Modified read macro to improve consistency between Windows and Linux builds, as well as c-embed
Modified Linux fopen to use binary mode, improves consistency and fixes c-embed usage
Remove webview/ path from WebView2.h include declaration
In _webui_http_handler(), first check if the path is a file to fix c-embed usage which does not properly support folder existence

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -40,7 +40,9 @@ ARCH_TARGET ?=
 # BUILD FLAGS
 CIVETWEB_BUILD_FLAGS := -o civetweb.o -I"$(MAKEFILE_DIR)/include/" -c "$(MAKEFILE_DIR)/src/civetweb/civetweb.c" -I"$(WEBUI_TLS_INCLUDE)" $(TLS_CFLAG) -w
 CIVETWEB_DEFINE_FLAGS = -DNDEBUG -DNO_CACHING -DNO_CGI -DUSE_WEBSOCKET $(TLS_CFLAG)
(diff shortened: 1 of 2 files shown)


nater0000:
WebView2.h was not removed, just the “webview/“ folder name is removed from
the include. I think this will help other projects as well, since
WebView2.h does not need to be within the webview directory.

c-embed is not required for anything; this change just supports filesystems
where folders may not exist or be reliably accessible. There is no c-embed
dependency.

webui_get_port() should have been added in webui.h and a similar function
can be added to webui.hpp

My apologies if I am misunderstanding any of these reasons.

nater0000:
No, I am also using a local copy of WebView2.h, but I have added “webview”
into the list of Include Directories for my local build, since I do not use
webui’s Makefile.
Not every project that builds webui.c may want to maintain this
“webview/WebView2.h” directory structure.
Is there a reason not to add “webview” into the build’s list of include
paths (instead of hard-coding the include to require WebView2.h to be
located within a directory named “webview”?

nater0000:
#include “WebView2.h” is more flexible than #include “webview/WebView2.h”
and should likely be preferred unless there is a reason to require that
specific folder structure.
It does not need to be part of this PR, but I suggest adding the change for
any other projects that want to build webui

If preferred, I can create a new PR without this change and any others that
may interfere with merging but are not critical for my use case (checking
if the requested path is a file before checking if the path is a folder).

Thank you!

PR #448, opened 2024-07-09
exposing _webui_get_free_port

sensiblearts:
I want to select a port dynamically for my custom web server, and I figured that since the code is already in `webui.c`, just expose it in the headers.

(If this PR is accepted, I have a `zig-webui` PR with an example app.)


sensiblearts:
I added the wrapper and pushed the changes.

PR #455, opened 2024-07-12
feat: add handlers for javascript alert, confirm & prompt.

fernandolguevara:
- alert modal
image

- confirm modal
image

- prompt modal
image


fernandolguevara:
@hassandraga we can't make this functionality with javascript... It's necessary the underlying layer take care of this using WKWebview on macos

PR #487, opened 2024-09-30
add setting of minimum window size & default user icon patching. (currently windows only)

testdrive-profiling-master:
add setting of minimum window size. (currently windows only)
-> I hope someone will make an implementation for linux and mac.
add default user icon patching. (currently windows only)
-> Conventionally, the first icon(resource ID : 101) in Windows is the main icon.


testdrive-profiling-master:
.editorconfig & 'windows.h' are deleted.
I will keep this request! ;)

PR #488, opened 2024-09-30
File handlers could receive window number

vieiro:
This is the proposed implementation for the discussion at https://github.com/webui-dev/webui/discussions/486.


vieiro:
@AlbertShown so do we want an example for `webui_set_file_handler_window` too? I can create one as well...

vieiro:
Updated the `@example` for webui_set_file_handler_window`!

vieiro:
D'oh! Very true! I added the `@example` to the wrong comment.
Do you want me to squash all commits together?

PR #491, opened 2024-10-07
fix correct window size from webui client size. (windows)

testdrive-profiling-master:
I found 'AdjustWindowRect()' function to make correct window size for windows.


testdrive-profiling-master:
Yes, I request newly merge PR #494.

2024년 10월 9일 (수) 오후 10:23, Showns ***@***.***>님이 작성:

PR #495, opened 2024-10-10
Fix release mode building with Zig

mochalins:
The `build.zig` linker flags setting caused issues with linking `civetweb`, notably redefinition of `MUST_IMPLEMENT_CLOCK_GETTIME` macro and a warning for using the `__DATE__` macro (which turns into an error with default Zig compiler flags). This occurred when attempting to build `webui` with Zig in any release mode on Windows (potentially other OSes as well, but untested).

This patch addresses these issues and allows for building on release modes in at least Windows and Linux systems.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -68,9 +68,8 @@ fn addLinkerFlags(b: *Build, webui: *Compile, enable_tls: bool) !void {
     // Prepare compiler flags.
     const tls_flags = &[_][]const u8{ "-DWEBUI_TLS", "-DNO_SSL_DL", "-DOPENSSL_API_1_1" };
     var civetweb_flags = std.ArrayList([]const u8).init(b.allocator);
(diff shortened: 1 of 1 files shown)


wwderw:
This happens on Linux as well.  Work around that I had was just reverting back to zig .13 when building webui.

mochalins:
> This happens on Linux as well. Work around that I had was just reverting back to zig .13 when building webui.

Does this patch fix your issue? I have self-hosted CI running `ReleaseSafe` builds for Fedora Linux 40 (kernel 6.6.12) successfully with this patch, but haven't tested on any other Linux distros.

wwderw:
> > This happens on Linux as well. Work around that I had was just reverting back to zig .13 when building webui.
> 
> Does this patch fix your issue? I have self-hosted CI running `ReleaseSafe` builds for Fedora Linux 40 (kernel 6.6.12) successfully with this patch, but haven't tested on any other Linux distros.

Using 1827 build, compiling to x86_64-linux-musl ReleaseSmall, it worked just fine (on Linux, I tend to build for Musl, so that's the begin one for me).

jinzhongjia (approved):

PR #501, opened 2024-10-12
Zig build improvements

mochalins:
Clean up and improvements to current Zig build process. Notably:

1. Specify paths in `build.zig.zon` to limit files stored in cache when `webui` is fetched as a dependency
2. Set `verbose` option to be an enum of levels, so that more specific build information can be emitted 
3. Fix bug where prior `verbose` option printed `Done.` immediately before build process completed
4. General cleanup of linker flags and file paths to adhere to Zig best practices

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -6,9 +6,9 @@ const OptimizeMode = std.builtin.OptimizeMode;
 const Compile = Build.Step.Compile;
 const Module = Build.Module;
 
-const log = std.log.scoped(.WebUI);
 const lib_name = "webui";
 const zig_ver = builtin.zig_version.minor;
+var global_log_level: std.log.Level = .warn;
 
 pub fn build(b: *Build) !void {
     const target = b.standardTargetOptions(.{});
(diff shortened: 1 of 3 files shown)


build.zig
```diff
@@ -118,34 +134,56 @@ fn build_examples(b: *Build, webui: *Compile) !void {
 
     const examples_path = (if (zig_ver < 12) (Build.LazyPath{ .path = "examples/C" }) else b.path("examples/C")).getPath(b);
     var examples_dir = if (zig_ver < 12)
-        try std.fs.openIterableDirAbsolute(examples_path, .{})
```
jinzhongjia, line 121: Is there any difference between these two usages?
Is there any relevant documentation or discussion notes?

jinzhongjia (approved):

build.zig
mochalins: The Zig core team recommends avoiding absolute paths in all situations, as numerous issues have been filed in the past related to absolute file paths used in very unusual mounting/filesystem environments. 

As such, Zig best practice is to pin all filesystem operations on `cwd` directory base. Absolute file paths even when operated from `cwd` directory will still resolve successfully, so there is no reason not to follow this best practice.

As for documentation/discussion notes, here's one relevant curently open issue:

jinzhongjia: I see
This is not the same as my past experience using c

PR #508, opened 2024-10-24
Fix memory error

jimying:
- mismatch error,  use strdup() alloc memory, but use _webui_free_mem() to free
- double free error,  _webui_get_full_path() already free 'file' internal, but free again when use
- can't free ws library memory data which not alloced by webui
- can't free const memory


mochalins (approved):
All of these are great fixes, thank you also for the descriptions in your PR!

PR #516, opened 2024-11-03
make sure 4096 bytes not alloc 8192 bytes

kbkpbot:
Make sure calling `web_malloc(4096)` will alloc 4096 bytes indeed, not 8192 bytes.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -3755,7 +3755,7 @@ static size_t _webui_mb(size_t size) {
     size--;
 
     size_t block_size = 4;
-    while(block_size <= size)
+    while(block_size < size)
         block_size *= 2;
 
     return block_size;


mochalins (approved):

mochalins:
Note: It is preferable to keep PRs that address different issues separate. It'll make it easier to merge, and also potentially revert isolated issues in the future.

mochalins:
@hassandraga is there a reason this was reverted?

mochalins:
> Deno was crashing while I was working on https://github.com/webui-dev/deno-webui/commit/4459765896a67aacd0b5d1d4edc3af867c523351
> 
> I did not deeply investigate yet, but reverting this solved the issue.

Do you possibly have a repro/example project posted somewhere? I'd love to debug this with Deno as well. This fix seems quite highly likely to be issue-free, and I suspect some other bug was being covered up by the incorrect larger allocations.

PR #517, opened 2024-11-03
add mutex for malloc/free

### LINKED ISSUE #507: data race when memory alloc/free (ws non-blocking mode)
In  ws non-blocking mode (win->ws_block = false),
WEBUI_WS_DATA  will process in new thread (_webui_ws_process_thread),  which call _webui_free_mem() to free memory.
If muti-threads free and main thread _webui_malloc() to alloc memory, maybe cause  data race error (may crash).

we should add a mutex lock.

kbkpbot:
1. add `mutex_mem`, the will fix #507 
2. remove `0x` prefix from all `%p`, as `%p` already has one `0x` prefix
3. merge PR #516

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -395,6 +395,7 @@ typedef struct _webui_core_t {
     webui_mutex_t mutex_webview_stop;
     webui_mutex_t mutex_http_handler;
     webui_mutex_t mutex_client;
+    webui_mutex_t mutex_mem;
     webui_condition_t condition_wait;
     char* default_server_root_path;
     bool ui;
@@ -770,7 +771,7 @@ bool webui_script_client(webui_event_t* e, const char* script, size_t timeout,
     #ifdef WEBUI_LOG
     printf("[User] webui_script_client([%zu])\n", e->window);
     printf("[User] webui_script_client([%zu]) -> Script [%s] \n", e->window, script);
-    printf("[User] webui_script_client([%zu]) -> Response Buffer @ 0x%p \n", e->window, buffer);
+    printf("[User] webui_script_client([%zu]) -> Response Buffer @ %p \n", e->window, buffer);
     printf("[User] webui_script_client([%zu]) -> Response Buffer Size %zu bytes \n", e->window, buffer_length);
     #endif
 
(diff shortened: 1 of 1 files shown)


AlbertShown:
> 1. add mutex_mem, the will fix https://github.com/webui-dev/webui/issues/507

✅ Mutex codes is great, Thank you 👍

✅ Yes, using `ARRAY_SIZE` is better than `WEBUI_MAX_IDS` in memory pointers array 👍
The reason behind `WEBUI_MAX_IDS` is to know how many windows and server threads we can have.

❌ But I don't know what's the problem you are trying to solve by modifying `_webui_ptr_add()` loop... the mutex code you added is indeed solves the data race issue, but modifying the whole loop it's not clear to me why.

> 2. remove 0x prefix from all %p, as %p already has one 0x prefix

✅ In Windows I guess some compilers does not add it, however, don't have it at all is better than have it twice. So, Thank you 👍

> 3. merge PR https://github.com/webui-dev/webui/pull/516

The issue is the whole webui code does not have the `malloc(buf_size + 1)` style, but just `malloc(webui_mb(buf_size))`.
The reason is when a developer asks the OS to allocate 13 bytes for example, the OS's kernel allocate a full block of memory, like 16, 32, 64, 128... so it's recommended to allocate a full block instead, with + one extras byte for strings.

On the other hand, probably this memory block thing is an overkill, maybe we don't need it in today's modern systems... How about we just add the extras one byte and that's it?

```c
static size_t _webui_mb(size_t size) {
    // Make sure we have the null terminator byte for strings data type
    return size++;
}
```

AlbertShown:
> I think the global value _webui.ptr_position , _webui.ptr_list, _webui.ptr_size need to be protected by mutex, so I add mutex protect them when accessing these values;

Yes 👍

> I rewrite _webui_ptr_add() because I just want to save the code
> Or we just mutex the whole _webui_malloc, _webui_free...As the loop will not take much time;

Yes, It's a small, tiny loop, it won't take time. I suggest we keep the loop as it is.

> I think totally remove _webui_mb() is a good choice. This has been done by memory allocator already. For strings, I think the programmer need to know that it will take one extra byte save the zero, so he will call webui_malloc(strlen(s)+1) instead of webui_malloc(strlen(s)).

Removing `_webui_mb()` completely will force us to update 114 line, I suggest we just keep it, but we make it sample. And it here if we need to do something with buffers size in future.

```c
static size_t _webui_mb(size_t size) {
    // Make sure we have the null terminator byte for strings data type
    return size++;
}
```

mochalins:
I'm going to split 3. back into its own PR, as it is a much simpler merge on its own.

Issue #518, opened 2024-11-07
ThreadSanitizer data race inside webui_show_browser


Lev275568:
revision d5b25ca1a827d3ec07be9607232edb0c27172918

```
void sandbox()
{
    size_t handle = webui_new_window();
    assert(handle != 0);
    assert(webui_set_root_folder(handle, "/home/user2"));
    webui_show_browser(handle, "unknownFile_ieshoi3G.html", Chromium);
    std::abort();
}
```

> 
> WARNING: ThreadSanitizer: data race (pid=154801)
>   Read of size 4 at 0x7ffff4427028 by thread T1:
>     #0 master_thread_run /webui/src/src/civetweb/civetweb.c:20466:9 (test+0x1910956) 
>     #1 master_thread /webui/src/src/civetweb/civetweb.c:20593:2 (test+0x18f5402) 
> 
>   Previous write of size 4 at 0x7ffff4427028 by main thread:
>     #0 mg_stop /webui/src/src/civetweb/civetweb.c:20722:2 (test+0x18f1cff) 
>     #1 _webui_socket_test_listen_mg /webui/src/src/webui.c:4088:5 (test+0x18da4b6) 
>     #2 _webui_port_is_used /webui/src/src/webui.c:4106:10 (test+0x18d8cf5) 
>     #3 _webui_get_free_port /webui/src/src/webui.c:7423:17 (test+0x18d8bd2) 
>     #4 _webui_show_window /webui/src/src/webui.c:7005:56 (test+0x18dac29) 
>     #5 _webui_show /webui/src/src/webui.c:6728:16 (test+0x18d8317) 
>     #6 webui_show_browser /webui/src/src/webui.c:1673:12 (test+0x18d8121) 
> 
>   Location is heap block of size 165368 at 0x7ffff4427000 allocated by main thread:
>     #0 calloc /compiler-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:686:5 (test+0x82687b) 
>     #1 mg_calloc /webui/src/src/civetweb/civetweb.c:1497:9 (test+0x18ee9a5) 
>     #2 mg_start2 /webui/src/src/civetweb/civetweb.c:20979:29 (test+0x18f231a) 
>     #3 mg_start /webui/src/src/civetweb/civetweb.c:21700:9 (test+0x18f5489) 
>     #4 _webui_socket_test_listen_mg /webui/src/src/webui.c:4078:16 (test+0x18da487) 
[...]

PR #522, opened 2024-11-18
Allow for building `civetweb` in debug mode through `build.zig`

mochalins:
As #518 shows, it seems useful to be able to build the vendored `civetweb` dependency in debug mode as well.


mochalins:
> If this will make civetweb's logs being printed and mixed with webui's logs, I suggest adding a specific extras argument to enable civetweb logs.

Haha actually webui logging is not printed in any mode right now! But I do agree that a flag to differentiate/conditionally enable civetweb logs would be helpful. I'll make the change shortly.

mochalins:
This should be rebased after #525

PR #579, opened 2025-04-08
Drag drop compatibility

konakona418:
Used a CSS custom property to improve compatibility with `-webkit-app-region`.

This patch introduces a new CSS property `--webui-app-region`, which can be used by setting `--webui-app-region: drag;` in the element's style. On platforms that support `-webkit-app-region`, the original behavior remains unaffected.

This approach helps ensure broader compatibility across environments while preserving the intended drag behavior.

For more context, please refer to the discussion in PR #577.

> I've done some basic testing, but feel free to double-check if everything works as expected. Also, please verify that the TypeScript code is properly transpiled. Thanks!


konakona418:
Thanks!

Issue #276, opened 2023-11-29
[Feature Request] Add Conan package


redradist:
It would be much easier to use conan in C/C++ environment through package

PR #586, opened 2025-04-16
Update CMake build system and introduce Conan and VCPKG

Juff-Ma:
This updates the CMake build system to have more features and introduces both [Conan](https://conan.io) and [VCPKG](https://vcpkg.io) in order to support easier fetching of dependencies (OpenSSL and WebView2)

The changes made to the CMake itself are:
- Added metadata like version and description
- Explicitely defined C version as 99
- Add `WEBUI_LOG` and `NDEBUG` flags depending on configurarion (like build.zig or makefiles)
- Add the 'lib' prefix on macOS and Linux (parity with makefile behaviour)
- Allow the user to change the output library name (webui by default uses different names, e.g. -static, -secure, when building, this allows the user to remove those if desired)
- Fix dll not getting copied when installing (added runtime install directive)
- Add support for building ObjC on macOS (webview)

### Conan and VCPKG
Conan and VCPKG are both package managers used for C and C++. Here I included them to automatically download and build OpenSSL, this makes it trivial to build webui with OpenSSL support without having an OpenSSL dev environment installed.

Using Conan it's just `conan build . -of build -o tls=True`. If Conan and a compiler environment are installed this will do the following:
- Download CMake
- Download and build OpenSSL
- Configure webui based on Conan options
- Build and link webui

With VCPKG the process is more complicated but you also have more granular control about it (e.g. cross compiling), it more or less boiles down to

1. `vcpkg install --x-feature=tls` which will download and build OpenSSL
2. `cmake -B build -S . -DCMAKE_TOOLCHAIN_FILE=<vcpkg_install_dir>\scripts\buildsystems\vcpkg.cmake -DWEBUI_USE_TLS=ON`
3. `cmake --build build`

This will result in more or less the same result as the Conan command.
Additionally VCPKG will download WebView2, however, since webui doesn't link against it but rather load the dll dynamically the user still needs to copy it manually, changing this _could_ probably be possible using a newer CMake version but I didn't look into it too deep.

Note that neither Conan nor VCPKG are included in their package forms but only for building webui itself. This is because Conan's package format is (in my opinion) a bit confusing and VCPKG's packages are mostly managed in a central repo by Microsoft (which already includes a webui package).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,7 @@ CMakeUserPresets.json
 **/conan-*
 **/cmake-build-*
 **/deactivate_conanbuildenv*
+
+# vcpkg
+vcpkg-manifest-install.log
+vcpkg_installed/
--- a/.gitignore
+++ b/.gitignore
@@ -125,4 +125,11 @@ Thumbs.db
 **/cmake_install.cmake
 **/CMakeCache.txt
 **/CMakeFiles/
+CMakeUserPresets.json
 .zig-cache
+
+# Conan
+**/compressor
+**/conan-*
+**/cmake-build-*
+**/deactivate_conanbuildenv*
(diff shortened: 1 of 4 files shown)


Juff-Ma:
> Thank you so much @Juff-Ma, this is very helpful 👍 Your PR also solves #276.

I don't know about that. Like I said it does not include the packaging. However it should be easy to implement, the docs are just a bit all over the place.

Juff-Ma:
I think the functionalty of `WEBUI_OUT_LIB_NAME` will prove useful in this case. This should simplify the copying and linking information.

PR #614, opened 2025-07-18
Check webui.js load hostname and connect to it.

MiyamuraMiyako:
`window.location.hostname` get hostname from main url, but if main url's hostname is not localhost, will can't load webui.js. Use stacktrace to find the actual url of js
图片


MiyamuraMiyako:
Oh, It seems still has issue.
图片

MiyamuraMiyako:
> Oh, It seems still has issue. 图片

Ok, I found this config: `webui::set_config(multi_client, true);`

MiyamuraMiyako:
图片
图片

I tested the program on the aarch64 emulator of x64 pve, and the performance was not very high. The 1.5 second wait does not seem to be enough to complete the loading, maybe it should be set longer here?

MiyamuraMiyako:
It seems that some `alert` are only suitable for testing. Wouldn’t it be better to use `console.log` instead?

AlbertShown:
Thank you for the PR @MiyamuraMiyako.
But throwing an error to get the stack is a hack more than a stable code, it works but is inefficient and not guaranteed across all environments. Also throw `new Error()` generates a error stack trace.

1. We should find a better stable way
2. The code should also support `https` when using webui-2-secure.dll

AlbertShown:
> It seems that some alert are only suitable for testing. Wouldn’t it be better to use console.log instead?

Yes! I agree with you 👍
a PR for that is welcome.

AlbertShown:
> The 1.5 second wait does not seem to be enough to complete the loading, maybe it should be set longer here?

I taught somebody already pushed a PR to change this to `3000` (_3 seconds_).
However, a PR is welcome.

MiyamuraMiyako:
> Thank you for the PR @MiyamuraMiyako. But throwing an error to get the stack is a hack more than a stable code, it works but is inefficient and not guaranteed across all environments. Also throw `new Error()` generates a error stack trace.
> 
>     1. We should find a better stable way
> 
>     2. The code should also support `https` when using webui-2-secure.dll

Ok, I will check `document.scripts` to find webui.js reference.
And I think no one will put `http://.../webui.js` in case of https service?

MiyamuraMiyako:
> > It seems that some alert are only suitable for testing. Wouldn’t it be better to use console.log instead?
> 
> Yes! I agree with you 👍 a PR for that is welcome.

OK, I will submit a new PR later.

MiyamuraMiyako:
I will open new pr.

PR #618, opened 2025-07-24
Check webui.js load hostname and connect to it.

MiyamuraMiyako:
`window.location.hostname` get hostname from main url, but if main url's hostname is not localhost, will can't load webui.js. Use `document.scripts` to find the actual url of js
图片

Then add c++ method `set_browser_folder`, export from c function `webui_set_browser_folder`.


MiyamuraMiyako:
@AlbertShown I think everything is ready to merge.

AlbertShown:
Thank you @MiyamuraMiyako for your contribution

PR #623, opened 2025-07-30
increase websocket connection timeout

sigmaSd:
The current timeout failed many times on my pc, 5 seconds seems to be safer

Also if it fails, some times it segfaults at exit, seems like a double free somewhere, I tried following the trace with gdb but its not straight forward, you can reproduce if you make the time out 0 and try to open it many times

ref https://github.com/webui-dev/deno-webui/issues/64#issuecomment-3127094542


AlbertShown:
Yes, I think 1.5s is too short, and 10-30s maybe too long, so 5s looks better.

AlbertShown:
I set timeout to `0` and I tested it with the C examples, and also with Deno examples, I can't reproduce the segfault issue. However, if increasing the timeout from `1500` to `5000` solved your issue, then all fine.

PR #625, opened 2025-08-02
Win32 - Fix uninitialized vars

swairr:
Fix uninitialized variable on Windows.

`wc.lpszMenuName` is not assigned, and will crash when `RegisterClassA`.


AlbertShown:
Thank you @swairr

PR #636, opened 2025-08-31
Two changes:

hdijkema:
1. Making sure window moves will be visible in window.outerWidth/outerHeight (only webview2(?)). It seems the only way to do that is to do some resizing on the window if it is moved. See MOVE: switch. 
2. Make it possible to register a custom log handler, so that in debug mode, logging can be assembled along with other application logs in the application webui is embedded in.

### DIFF
--- a/include/webui.h
+++ b/include/webui.h
@@ -206,6 +206,19 @@ typedef struct webui_event_t {
     char* cookies;          // Client's full cookies
 } webui_event_t;
 
+// -- Other types ---------------------
+
+// custom log handler type
+typedef enum {
+    log_debug = 1,
+    log_info,
+    log_warning,
+    log_error,
+    log_fatal
+} webui_log_level_t;
+
+typedef void(*webui_custom_log_handler_t)(webui_log_level_t level, const char *);
+
 // -- Definitions ---------------------
 
 /**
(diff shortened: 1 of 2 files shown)


AlbertShown:
Thank you for this PR. I need to test both to see custom log behavior in release version (no logs). I think close handler will work. However, maybe better if each one has a separated PR but I can test both and marge for now.

hdijkema:
> Thank you for this PR. I need to test both to see custom log behavior in release version (no logs). I think close handler will work. However, maybe better if each one has a separated PR but I can test both and marge for now.

How can I make two different PRs? (for future PRs).

And maybe related, maybe not. When I close the last window (using webui_close), the next time I create a new window, it will get a 0 as number, i.e. it fails to create. Is this desired behaviour?

hdijkema:
One other thing, see commit https://github.com/webui-dev/webui/pull/636/commits/9c16b4acb7a9c84748c7399c6b7a152cbc6bb0a7

I'm not able to test it on MacOS, Linux, I can do.

This change ensures that when the last window closes, the (DLL) library is cleaned up and can also be re-initialized.

_Why is this important?_

Some environments (like the DrRacket IDE) keep DLLs loaded (even when one requests to unload them).

After closing the last window, the in the current implementation of webui, the DLL stays in 'exit' state and it is not possible anymore to create new windows.

This change makes sure that it is possible to keep the shared library working while loaded. 
A call to webui_clean(), will make sure that the library also reinitializes on calls to _webui_init() and doesn't stay in 'exit' state. 

In an FFI one can keep administration of opened / closed windows and at the last closed window call 'webui_clean()'.

AlbertShown:
The commit https://github.com/webui-dev/webui/commit/9c16b4acb7a9c84748c7399c6b7a152cbc6bb0a7 looks handy, thank you for that. It should be in a separated PR to be merged.

The other two changes, `custom log` and `close handler` should be tested first before merge, also preferable in a separated new PRs.

hdijkema:
I've worked on WebUI on Linux. 

- Made sure it can be linked against a Gtk-application (hence the -fPIC and static declarations).
-  Made available the native handle for the Gtk WebView solution.

AlbertShown:
Thank you @hdijkema for this PR, really appreciated. However, this PR modifies `+826 −583` lines, which is huge, and I think this is like 2 or 3 fixes.

To better tracking the changes, I really which if you can create new separated PRs for each one.

1. Clone latest `main` branche
2. Do one modification, for example this one https://github.com/webui-dev/webui/commit/9c16b4acb7a9c84748c7399c6b7a152cbc6bb0a7
3. Push a new PR with a title like `Making webui re-usable after cleaning`
4. Do another modification, for example `logs`...
5. Push a new PR...

Again, thank you @hdijkema for all those fixes.

hdijkema:
Hi AlbertShown,

I'll do this some time soon. 
There's more coming up, because also on Linux, GtkWebKit doesn´t generate an event for navigations without some signal connecting.

AlbertShown:
@hdijkema in future, if you create new PRs, please try to avoid modifying `webui.h` as possible because that will make all wrappers need to be updated as well. Also, a PR should solve one specific issue.

PR #637, opened 2025-10-05
Add CMake option to control building of examples

kaniteru:
Hello,

I encountered an issue when using `FetchContent` to include WebUI in another CMake project.
In this case, `CMAKE_SOURCE_DIR` does not point to the WebUI root directory, which results in errors.

Additionally, the example projects are always built, even when they are not needed.

So, I made a few small changes to fix this:
- Replaced `CMAKE_SOURCE_DIR` with `CMAKE_CURRENT_SOURCE_DIR`
- Added a `WEBUI_BUILD_EXAMPLES` option to control whether example projects are built

Thank you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,9 @@ if (WEBUI_USE_TLS)
     set(WEBUI_DEFAULT_OUT_LIB_NAME "webui-2-secure")
 endif()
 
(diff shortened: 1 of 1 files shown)


AlbertShown:
Thank you @kaniteru for the CMake improvements.

PR #639, opened 2025-10-13
Fix overwritten event entries in _webui_new_event_inf

eric-boyd-protolabs:
When creating new event info data and number, search for an empty event array entry rather than simply increasing the number and potentially overwriting existing info. This prevents intermittent double deletions when event blocking is disabled and many events of varying run durations occur in a short time.

Also increase the number of event entries that can be used to WEBUI_MAX_IDS which is the allocated events array size.

Note that this change can introduce a deadlock (rather than a double delete) when the number of active events is greater than the event array size due to the introduction of the lock. That could probably be handled by introducing a event entry check counter and outer loop with a wait while the counter matches WEBUI_MAX_IDS. _webui_new_event_inf would block until an open slot became available.


AlbertShown:
Thank you @eric-boyd-protolabs for the fix.

PR #641, opened 2025-10-17
Fix a problem with WebView2, which won't generate a move event

hdijkema:
Fix a problem with WebView2, which won't generate a move event
when the window is moved. This is done by simulating a resize for WebView2, 
which will trigger the move event eventually.


AlbertShown (approved):

AlbertShown:
Thank you @hdijkema

PR #642, opened 2025-10-17
Add functionality to use a custom log handler

hdijkema:
Add functionality to use a custom log handler, which facilitates
the integration of webui in software. Debug logging of webui 
can now be integrated in logging systems of software that uses webui.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -214,28 +214,28 @@ typedef struct webui_event_inf_t {
     typedef void (*gtk_window_set_position_func)(void *, int);
     typedef unsigned long (*g_signal_connect_data_func)(void *, const char *, void (*callback)(void), void *, void *, int);
     // GTK Symbol Initialization
-    gtk_init_func gtk_init = NULL;
-    gtk_widget_show_all_func gtk_widget_show_all = NULL;
-    gtk_main_iteration_do_func gtk_main_iteration_do = NULL;
-    gtk_main_func gtk_main = NULL;
-    gtk_main_quit_func gtk_main_quit = NULL;
-    g_timeout_add_func g_timeout_add = NULL;
-    gtk_events_pending_func gtk_events_pending = NULL;
-    gtk_container_add_func gtk_container_add = NULL;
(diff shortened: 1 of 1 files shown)


AlbertShown:
I totally see the point here, but I'm wondering if this is necessary, especially the fact that all other wrappers need to implement this new API and will be available only in `debug` version of WebUI. In general people does not use debug version except people who actually developing core code like you, so we won't necessarily need this logging API to debug WebUI core code.

Please let me know if I miss something.

hdijkema:
Well, I've been debugging quite a lot to see what's going on and why some events work and some not in what circumstances. This is while working on a Web application. E.g., why don't I get a 'move' event, or how is it possible I don't get a navigation event. Or why does GtkWebView crash? Or to understand why things "lock up". Without the debug information that's a stretch to do. 

Look e.g. at programs like rsync, openvpn, etc. they all have a flexible logging system with more or less detailed information about what's going on under the hood. It's a good practice, especially to understand the behaviour of the software. 

In larger software projects people tend to create some logging functionality, e.g. Log4j, log4c. Also lots of programs log to syslog, or trace to a file. Of course one could redirect stdout and stderr to a file, but this is not always easy to do. E.g. my IDE catches stderr/out and the library output (of webui) just "disappears". 

I think this code is not intrusive, i.e. when no log handler is installed, it will just log to stdout/stderr. And when one chooses to install a log handler, it will redirect logging to the handler. Even if someone chooses not to implement the call to the log handler installer for a wrapper, it's not really a problem. But someone might choose to implement it, because of the reasons above. 

In the end, it's your project, so the decision is yours to make.

AlbertShown (approved):

AlbertShown:
Fair enough @hdijkema, I was just trying to make minimal change as possible in `webui.h` cause that force lot of other wrappers to be updated as well. I think this PR now have some code conflicts needs to be solved first.

hdijkema:
Ai. I'll try to resolve the conflicts, otherwise all discard this pull request and make another.

PR #643, opened 2025-10-17
"On Close" events and handling the window close for WebView solutions

hdijkema:
Functionality to callback to the software that integrates webui,
on the event of clicking a 'close' button. When the user tries to close the window, 
using WebView on Windows/Linux, webui will issue a callback (if installed) to the 
integrating software asking if the window may be closed.

Not created for MacOS WebView currently, as I have no access to MacOS

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -8,8 +8,6 @@
   Canada.
 */
 
-#define webui_log_debug printf
-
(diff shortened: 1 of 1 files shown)


AlbertShown:
Thank you @hdijkema, good feature. We should check out macOS later 👍

PR #646, opened 2025-10-17
Fix problems with navigation events that aren't generated by GtkWebKit.

hdijkema:
Fix problems with navigation events that aren't generated by GtkWebKit.
(used lib: libwebkit2gtk-4.1-0 on linux Mint).

The WebKit engine to apparently doesn't catch the navigations and does not produce navigation events, when something like "webui_bind(_webui_win, "", webui_event_handler);" is called. Although this should be expected behaviour.

By implementing a handler for the GtkWebKit decision signal, this navigation event can still be generated.


AlbertShown:
Why did you close this PR?

hdijkema:
Problems with strlcpy. I see that it doesn´t compile on all linux distro's your checking, so I'm replacing it with strncpy.

PR #648, opened 2025-10-17
Fix problems with navigation events that aren't generated by GtkWebKit.

hdijkema:
Fix problems with navigation events that aren't generated by GtkWebKit.
(used lib: libwebkit2gtk-4.1-0 on linux Mint).

The WebKit engine to apparently doesn't catch the navigations and does not produce navigation events, when something like "webui_bind(_webui_win, "", webui_event_handler);" is called. Although this should be expected behaviour.

By implementing a handler for the GtkWebKit decision signal, this navigation event can still be generated.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -11840,7 +11840,7 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpReserved) {
 
                     _webui_window_event(
                         win, // Event -> Window
-                        connection_id, // Event -> Client Unique ID
+                        0, // Event -> Client Unique ID | This is a WebKitGTK Event, so we don't have any WebSocket client unique ID
                         WEBUI_EVENT_NAVIGATION, // Event -> Type of this event
                         "", // Event -> HTML Element
                         event_num, // Event -> Event Number


hdijkema:
There are some fails that do not seem to be related to the changes of this pull request.

AlbertShown:
> There are some fails that do not seem to be related to the changes of this pull request.

Yes, this is a Windows workflow fails to find TLS, this because GitHub added Windows Server 2025. I will try to fix it.

AlbertShown:
@hdijkema I noticed that catching events now have two different functions, one is the regular `webui_bind()` with empty string, and the other ones are those added in your recent PRs. To not force all wrappers to be updated, and to not force to update all documentation, I will try to redirect those WebView event to the regular one, so no wrappers and documentation will need to be updated.

hdijkema:
> @hdijkema I noticed that catching events now have two different functions, one is the regular `webui_bind()` with empty string, and the other ones are those added in your recent PRs. To not force all wrappers to be updated, and to not force to update all documentation, I will try to redirect those WebView event to the regular one, so no wrappers and documentation will need to be updated.

I'm curious how you want to do that. Oh I see. You can of course catch this in the regulare webui_bind() with an empty string.  Great. good Idea I think.

AlbertShown:
>I'm curious how you want to do that

I will use your own GTK code to catch event and sends it to `webui_bind()`, then will be blocked by default. Then user can use `webui_navigate()` API to perform the navigation. The goal is the same exact code that work for web browser should work also for WebView.

AlbertShown:
> There are some fails that do not seem to be related to the changes of this pull request.

Fixed in https://github.com/webui-dev/webui/commit/63fb9fcd037c35aaaafb0f81f30a6070f9d54ad1

PR #650, opened 2025-10-17
Fix problem with linking on Linux using static library and code relocation

hdijkema:
Fix problem with linking on Linux using static library and code relocation.

In order to be able to link on Linux to other shared libraries, 
while compiling e.g. a static webui library, the -fPIC is necessary. 
This has been tested on Linux.


hdijkema:
It looks like the windows build is blocking this merge. Not related to this pull request.

AlbertShown (approved):

PR #651, opened 2025-10-17
Array too small (MAX_BROWSER_FILES)

hdijkema:
Fixed problem with array allocation for executables array.
MAX_BROWSER_FILES needed to be 5. Also initialized all
array entries to NULL before assigning executable names.


hdijkema:
Nice, after so many years of C/C++ programming I still didn't know that char *a[N] = { 0 }  will initialize the complete array to 0.

AlbertShown:
Yes, it's a standard ANSI C since 1999 👍

hdijkema:
Working with C since 1988 😖

AlbertShown:
That's totally fine, and thank you @hdijkema for your recent fixes and improvements 👍

PR #652, opened 2025-10-18
Fix compilation on a windows target.

hdijkema:


hdijkema:
Whatever I do, it does not seem to influence anything in these windows workflows. I'm probably missing something. Does it even run cmake to create the makefiles for these builds?

PR #655, opened 2025-10-21
Fix endless recursion on webui_malloc in debug mode.

hdijkema:
Fix endless recursion on webui_malloc in debug mode.
We're sure the length of the debug message of _webui_malloc < 256, 
so in that case we have enough room to print the debug message 
without allocating memory with _webui_malloc, which prevents 
an endless recursion to _webui_log.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -739,7 +739,7 @@ static void _webui_log(size_t level, const char *format, va_list args) {
             return;
         }
         vsnprintf(buffer, needed_size + 1, format, args);
-        _webui_log_data.logger_func(level, buffer, _webui.logger_user_data);
+        _webui_log_data.logger_func(level, buffer, _webui_log_data.logger_user_data);
         if (buffer != buf) _webui_free_mem((void*)buffer);
     }
 }


AlbertShown:
> which prevents an endless recursion to `_webui_log`

Can you please show me where is that? because I can't find it.
if dynamic allocation failed then it will pass "_Memory allocation failed for log_" once only.

Another main reason is that `char buf[256];` not thread-safe approach, especially when `ui_event_blocking = false`.

hdijkema:
So, when _webui_malloc is called, _webui_malloc will log this:

afbeelding

Thus, the log handler gets called, which tries to allocate a buffer, which calls _webui_malloc, which debugs that it allocates memory, etc, etc, etc.  So, this fix effectively prevents _webui_malloc from being called, and because _webui_malloc only needs some 50 bytes or so to write it's debug message, there won't be any recursion.

AlbertShown:
I see, good catch. To fix this endless recursion and at the same time be thread-safe I suggest keeping dynamic allocation but using `malloc()` directly instead of `_webui_malloc`.

hdijkema:
That is fine by me, although I think stack allocation of a character array may have some advantages over heap allocation.

* It's less costly (no heap allocation/deallocation).
* It's faster to access (not really a point in this use case).

And allocating on the stack is afaik thread safe.

By using a local array of decent size, the majority of debug cases there's probably no malloc/free needed

AlbertShown:
> And allocating on the stack is afaik thread safe.

Sorry I read the modification quick, and I taught the buffer was `static` but it's not.

AlbertShown (approved):

PR #656, opened 2025-10-21
Fix for navigation with GtkWebKit, more in line with bind(<win>, "", <callback>).

hdijkema:
With this code change, the call to bind(<window>, "", <callback>),
will also catch the navigation event for GtkWebKit (linux WebView).

So no need anymore for an navigation handler, unless one would 
want to implement this as a feature for all webview / browser variants.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -11903,9 +11903,9 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpReserved) {
                     snprintf(buf, 20, "%d", navigation_type);
                     size_t nt_s = strlen(buf) + 1;
                     char *type = (char *) _webui_malloc(nt_s);
-                    strncpy(uri, buf, nt_s - 1);
-                    uri[nt_s] = '\0';
-                  
+                    strncpy(type, buf, nt_s - 1);
+                    type[nt_s] = '\0';
+
                     // Event Info
                     webui_event_inf_t* event_inf = NULL;
                     size_t event_num = _webui_new_event_inf(win, &event_inf);


AlbertShown:
I did it in a different way (https://github.com/webui-dev/webui/commit/13a9d5a3d4d49062257c0d553306503429b4215c).
Can you please test latest main Branche in Linux?

hdijkema:
> I did it in a different way ([13a9d5a](https://github.com/webui-dev/webui/commit/13a9d5a3d4d49062257c0d553306503429b4215c)). Can you please test latest main Branche in Linux?

Hi AlbertShown,

I've reviewed your code. No need for testing yet I'm afraid, because you're missing one point:

When one sets new content manually (with some show form), be it HTML or a URL or a file, GtkWebKit will also generate a decide-policy signal. If its going to be ignored,  the content will not be set, but the page will be cleared and the connection to the GtkWebKit window will be lost (i.e. no webui.js loaded). 

So you can´t always ignore the policy signal. Sometimes you must allow it. 

You can also not reliable trust the  navigation_type to be `WEBKIT_NAVIGATION_TYPE_OTHER`, because `WEBKIT_NAVIGATION_TYPE_OTHER` will also be the type if one sets `window.location.href` in  javascript. 

So a lot of corners and bends to be considered when catching this behaviour. 

If you look at my solution, you can see that whenever something is shown programmatically, a flag is set, which is reset in the navigation handler. This ensures that GtkWebKit loads the provided HTML/Url/File as one would expect when one sets the content explicitly.

hdijkema:
If you want, I can take your current work and work on it to get it working the way it should. Would that be a good idea?

hdijkema:
I've added some changes to your code to make it work. Closing this pull request.

AlbertShown:
> If you want, I can take your current work and work on it to get it working the way it should. Would that be a good idea?

Yes of course, that will be great, and thank you for that.
I'm not doing any macOS/Linux testing right now as I'm focusing on finishing some WebView2 issue on Windows.

hdijkema:
> > If you want, I can take your current work and work on it to get it working the way it should. Would that be a good idea?
> 
> Yes of course, that will be great, and thank you for that. I'm not doing any macOS/Linux testing right now as I'm focusing on finishing some WebView2 issue on Windows.

I've finished this change. See pull request #659.

PR #657, opened 2025-10-21
Fix bootstrap problem with logger function.

hdijkema:
Fix bootstrap problem with logger function.

To intercept this with a log handler, it must be possible to set 
the log handler before any logging takes place.

This logger_func / logger_user_data cannot be part of webui_core_t, 
because logging is already used in _webui_init() and _webui of
type webui_core_t gets reset to 0 in _webui_init().

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -444,8 +444,6 @@ typedef struct _webui_core_t {
     char* default_server_root_path;
     bool ui;
     char* custom_browser_folder_path;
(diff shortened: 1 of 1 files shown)


hdijkema:
Also va_end needs to complement va_start (otherwise this could result in a memory leak or other problems). 
It's required. So probably my mistake earlier.

AlbertShown (approved):

PR #659, opened 2025-10-22
Fix on main branch to make navigation events with the Gtk WebView (GtkWebKit) work as expected.

hdijkema:
With these changes, I think the navigation eventing works as one would expect it
to do. When one calls webui_show* manually all is loaded. Interaction from 
within the HTML by clicks, javascript, etc. is catched.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -277,9 +277,6 @@ typedef struct webui_event_inf_t {
         bool stop;
     } _webui_wv_linux_t;
 
-    #define GTK_SET_SHOW(win, status) (if (win->webView && win->has_all_events) win->webView->in_show = status)
-    #define GTK_IS_SHOW(win) ((win->webView && win->has_all_events) ? win->webView->in_show : true)
-
 #else
     extern bool _webui_macos_wv_new(int index, bool frameless, bool resizable);
     extern void _webui_macos_wv_new_thread_safe(int index, bool frameless, bool resizable);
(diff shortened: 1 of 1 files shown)


AlbertShown:
Thank you for the fix.
I modified the code:

- Moving this fix to Linux side since it only effect GTK WebKit
- Renaming Macros name, so people in future can easily know what it does
- Moving set `in_show` from `_webui_show()` to `_webui_show_wv()` Linux side since it only effect GTK WebKit
- Finally replacing macros with inline functions... (_just because GCC failed to build and I wanted to fix this quick_)

Please let me know if this modification does not break your fix.

hdijkema:
I'll look into it coming as soon as possible, maybe later this evening. Looks good though. Funny. I started with static inline functions instead of macro's, and went back to macro's and now you moved back to static inline functions. Why did gcc fail to build? I'm working with gcc on linux.  Mingw32?

AlbertShown:
Both are fine, we just pick whatever can be fixed faster. The errors are here:
https://github.com/webui-dev/webui/actions/runs/18722987935/job/53400253839

hdijkema:
Oke. Thanks. But your correction does not work. I'm looking into it now. See the stack of function calls and the place where you put the in_show status setters. I'm going to work out the place where they need to be to function properly. Hope that is oke for you. 

afbeelding

hdijkema:
> Both are fine, we just pick whatever can be fixed faster. The errors are here: https://github.com/webui-dev/webui/actions/runs/18722987935/job/53400253839

I think I see what's wrong there:

```c
(if (win->webView && win->has_all_events) win->webView->in_show = status)
```

may not be to gcc's  liking, because of the parentheses.

afbeelding

AlbertShown:
> Oke. Thanks. But your correction does not work. I'm looking into it now

Please feel free to update any part of the code as I'm editing the code without testing.

AlbertShown:
> may not be to gcc's liking, because of the parentheses.

Yes, definitely the parentheses, but we can keep the inline functions for now as both give the same results. It's just a question about where is the best place to call `_webui_wv_gtk_set_show()` and `_webui_wv_gtk_is_show()`.

hdijkema:
I'm sorry, I replaced them again with a #define. I'll try to explain my current solution in  the next pull request.

PR #660, opened 2025-10-22
So, I think this is a minimal solution to this GtkWebkit WebView navigation problem

hdijkema:
By debugging I learned that the webView structure is initially not created. 
Only further on in the process there will be a webView structure that can be checked for. 

Hence  the statement at line 12140 was only necessary, because before that there was no webView structure to set variable `in_show` to true. 

Because of this I moved the `in_show` attribute to the `_webui_window_t`. That will always be there.  Only when on linux though. 

Now the only place to set in_show to true is at the beginning of `_webui_show_window`, before all other processing begins.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -12338,8 +12338,6 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpReserved) {
 
         // Linux GTK WebView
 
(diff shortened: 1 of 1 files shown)


AlbertShown:
That's sounds perfect 👍
thank you @hdijkema

hdijkema:
> That's sounds perfect 👍 thank you @hdijkema

Thank you for letting me contribute.

PR #661, opened 2025-10-22
Fix for UTF-8 problems with the ascii printer.

hdijkema:
Characters need to be in the ascii range, otherwise invalid UTF-8 data can (and is) written.

So anything outside the 32-126 range is output as hexadecimal number.

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -666,7 +666,7 @@ static bool _webui_check_certificate(const char* certificate_pem, const char* pr
 #endif
 #ifdef WEBUI_LOG
 static void _webui_print_hex(const char* data, size_t len);
-static void (const char* data, size_t len);
+static void _webui_print_ascii(const char* data, size_t len);
 static int _webui_http_log(const struct mg_connection* client, const char* message);
 #endif
 static WEBUI_THREAD_SERVER_START;


AlbertShown:
I did those minor changes:

- Bring back the `"%c"` instead of `0x%02X` when byte is `0x00`. That's because we want to see logs like this `arg1¤arg2¤arg3` instead of `arg10xCFarg20xCFarg3`
- Moving your `register unsigned char c` up 👍

hdijkema:
Hi AlbertShown,

The problem with 0xCF in a UTF-8 context is that it is a marker for a multi byte sequence. 
So when an integrating language defaults to UTF-8, there will be interpretation, which can lead to exceptions.
Sorry for this side effect of the logging system.

For the sake of documentation:

_In UTF-8, multi-byte sequences are introduced by leading bytes that start with a specific pattern 
of ones followed by a zero, indicating the number of continuation bytes. Bytes that are part of 
a multi-byte sequence but are not the first byte are called continuation bytes, and they 
always start with the pattern 10xxxxxx._ 

_Leading bytes (head bytes):_

-    _2-byte sequence: 110xxxxx_
-    _3-byte sequence: 1110xxxx_
-    _4-byte sequence: 11110xxx_ 

So CF indicates a 2-byte sequence.

It might be better to do something else, e.g. a ~ (126) or a tab (9).

AlbertShown:
I see, I agree with you. I suggest we simply remove `0xCF` so it will be printed as hex.

AlbertShown:
Fixed in https://github.com/webui-dev/webui/commit/309305718c95754137b6e6a300bc62bc5b71abab.

Issue #664, opened 2025-10-28
New webview code for Windows - race condition / non sequential behaviour?


hdijkema:
I noticed today that when I display two HTML pages directly after another, e.g. http://127.0.0.1:xxxx/ and next, http://127.0.0.1:xxxx/c:/tmp/test.html, that they tumble over each other, as if they're executed in separate threads. 

Sometimes this leads to a long repeat (>1000x) of retrieving http://127.0.0.1:xxxx/favicon.ico + http://127.0.0.1:xxxx/, a kind of a race condition. 

Mostly it results in not loading test.html, but being left with what '/' gets as html. 

I didn't notice this behaviour until today. Could it be that the new webview2 implementation in C++ has something to do with it? 

Can webui_show's be sequentialized?

hdijkema:
I can add some new information:

So when I do a move or a resize of the window this triggers somehow a reload of the page in the Windows WebView:

```
00000080:init:stderr reopened for internal use to C:\Users\hans\AppData\Local\Temp\ujr8.0
00000080:init:stdout reopened for internal use to C:\Users\hans\AppData\Local\Temp\ujr8.1
00000027:LOG:WebWire Handler Started
00000054:MSG:Web UI Wire - v0.2.2 - (c) 2025 Hans Dijkema - MIT
00000046:MSG:Web Wire Http Server on http://127.0.0.1:0
new app
00000010:OK:new:1:1
00000018:MSG:_webui_win = 1
00000049:MSG:webui reports url : http://192.168.56.1:19880
00000027:MSG:Serving url path (1): /
00000041:MSG:Window 1 (1) connected - clientid = 0
00000057:MSG:web-ui-wire-handle-event: 1, web_ui_wire_handle_event
00000047:MSG:{"evt":"page-loaded","page_handle":1001001}
00000063:EVENT:page-loaded:1:{"evt":"page-loaded","page_handle":1001001}
00000038:MSG:Serving url path (2): /favicon.ico
00000040:MSG:Result of WebUI_Utils::waitUntil = 1

set-html 1 ../../../web-wire/test/test1.html
00000021:OK:set-html:1:2001001
00000054:MSG:the_file: C:\devel\racket\web-wire\test\test1.html
00000016:MSG:port = 19880
00000045:MSG:webui_get_url = http://192.168.56.1:19880
00000038:MSG:_base_url = http://127.0.0.1:19880
[...]

PR #665, opened 2025-10-31
I think this will solve issue  #664

hdijkema:
Simply not setting the navigate flag will probably help at that point.


AlbertShown:
That's true. The original C part was like below, no URL flag was set.

```c
webView->url = wURL;
webView->width = (win->width > 0 ? win->width : WEBUI_DEF_WIDTH);
webView->height = (win->height > 0 ? win->height : WEBUI_DEF_HEIGHT);
webView->x = (win->x > 0 ? win->x : (int)((GetSystemMetrics(SM_CXSCREEN) - webView->width) / 2));
webView->y = (win->y > 0 ? win->y : (int)((GetSystemMetrics(SM_CYSCREEN) - webView->height) / 2));
```

PR #667, opened 2025-11-09
Add bridge extension types

Mishieck:
I added the ability to extend the `WebuiBridge` with extension types. I added support for callback extensions. So, when callbacks have been bound on the backend, the types for those callbacks can be reflected on the frontend. The types can be added to the global scope using the following code:

```ts
import type { CallbackAttributes, WithExtensions } from "path/to/webui.ts";

declare global {
  export const webui: WithExtensions<{ callbacks: Callbacks }>;
}

type Callbacks = {
  init: CallbackAttributes<[], void>;
  dbGetOne: CallbackAttributes<[id: string], Uint8Array>;
  exit: CallbackAttributes<[], void>;
};
```

When any one of the callbacks is called, the name, arguments, and return value will be type-checked. The following code snippet shows valid function calls:

```ts
webui.call("dbGetOne", "item-id"); // Returns Promise<Uint8Array>
// or 
webui.dbGet0ne("item-id");
```

It would be helpful to publish the bridge as its own package on NPM or JSR so that it can be used easily on the frontend. This way, one could import it using a path like `@webui-dev/bridge` instead of writing a long path to the module.


AlbertShown:
> When any one of the callbacks is called, the name, arguments, and return value will be type-checked.

Thank you for the PR, but I still confused a bit, what do you mean exactly? Can you give a full complete example of use?

> It would be helpful to publish the bridge as its own package on NPM or JSR

No, it's not possible due to two main reasons, one is that `webui.js` is generated by WebUI core differently according to user's callbacks, random tokens, random IDs etc..., The second reason is that WebUI is a desktop app, so it may be used in an offline PC.

AlbertShown:
Right now, you can do:

```js
dbGetOne("item-id").then((response) => {
    // ...
});
```

Or 

```js
const response = await dbGetOne("item-id");
```

Mishieck:
Thank you for your response. An example of how to use this is [Deno WebUI Todo Vanilla](https://github.com/mishieck/deno-webui-todo-vanilla). In the provided repo, the bridge was copied to `./src/client/bridge` and the types were used in `./src/client/global.d.ts`.

The PR is meant to provide type checking for applications that use TypeScript for the frontend. Such applications compile to JavaScript before being used in the browser. Since the bridge is only used for types, none of the bridge code used for this purpose will be present in the final code. If the user used `@webui-dev/bridge` to import the types, it still will not matter because the package `@webui-dev/bridge` will be a dev dependency and will be completely removed in the desktop app.

AlbertShown:
Sorry @Mishieck I didn't back to this because I was still confused a bit.
However, this PR still in record in case.
Thank you.

Issue #673, opened 2025-12-25
Server mode does not serve index.html by default when accessing IP:PORT


triuk:
When running WebUI in server mode, it behaves like a web server. However, opening the root URL (IP:PORT) in a browser does not load the custom UI page. Instead, users must manually enter `IP:PORT/index.html`.

**Expected behavior:**
Navigating to `IP:PORT` should automatically serve `index.html`.

**Actual behavior:**
Opening `IP:PORT` results in a Not Found response, so the user has to access `IP:PORT/index.html` explicitly.

**Suggested fix:**
Default the server root to `index.html`.
In C, this is currently handled by:

```
const char* url = webui_start_server(myWindow, "/full/root/path");
```
I am doing cpp, so one workaround was overriding the start behavior manually, for example:
```
static std::string start_server(webui::window &win, const std::string &file)
{
    // Store start page
    detail::g_start_page = file;

    // Install VFS middleware
    win.set_file_handler(detail::vfs_router);

    // Extract native window ID (WebUI stores it as the first field)
    size_t id = *reinterpret_cast<size_t *>(&win);

    // ALLOW PUBLIC ACCESS
    webui_set_public(id, true);

    // Start server
    const char *url = webui_start_server(id, file.c_str());
    if (!url)
        return {};

    return std::string(url);
}

// Usage
start_server(window_, "index.html");
```

Key detail:
The default-page behavior depends entirely on the wrapper implementing VFS routing – e.g. in this wrapper:
```
static std::string g_start_page = "index.html";

static const void *vfs_router(const char *filename, int *length)
{
    if (strcmp(filename, "/") == 0) {
        static thread_local std::string tmp;
        tmp = "/" + g_start_page;
        filename = tmp.c_str();
    }
    return vfs(filename, length);
}
```

But for example, Python wrapper has no mechanism to override this, therefore default behavior must be implemented in WebUI core to be consistent across languages.

**Request:**
Add automatic serving of the configured start page when `/` is requested, so that entering `IP:PORT` loads the UI without requiring `/index.html`.

AlbertShown:
Thank you @triuk for pointing this out. Fixed https://github.com/webui-dev/webui/commit/6c8b11db5cb3e94616fdec1aed48deb4a3aadab8

triuk:
Could you please write an example, how to use it? Because no matter what I write to the cpp:
```
window_.set_file_handler(vfs);
std::string_view url = window_.start_server("index.html or whatever");
```
[...]

PR #681, opened 2026-01-28
Deterministic entry file handling for VFS and filesystem roots

triuk:
## Summary

This PR makes the entry file explicit and deterministic across all WebUI modes. It fixes `start_server()` with VFS so the requested entry file is respected, and it removes index auto‑detection/redirect behavior from the VFS generator. The result is consistent behavior for VFS and filesystem roots, in both server and non‑server modes.
Tested on Linux and Windows.

## What Changed

1. **VFS + `start_server()`** The VFS path requested by the application (e.g. `test.html`) is now honored rather than falling back to a hardcoded VFS index.
2. **VFS generator behavior (`vfs.py`)**
   Index auto‑discovery and folder redirect logic were removed. The VFS now serves only the exact path requested by the application.

## Testing Matrix and Behavior

```cpp
#include "webui.hpp"
#include "vfs.h"

int main(int argc, char *argv[])
{
    bool serverMode = false;
    bool vfsEnabled = false;
    for (int i = 1; i < argc; ++i)
    {
        if (std::string(argv[i]) == "--server")
        {
            serverMode = true;
        }
        else if (std::string(argv[i]) == "--vfs")
        {
            vfsEnabled = true;
        }
    }

    webui::window window_;
    webui::set_config(use_cookies, false);

    if (vfsEnabled)
    {
        window_.set_file_handler(vfs);
    }
    else
    {
        window_.set_root_folder("/path/to/ui/");
        // window_.set_root_folder("/path/to/ui");
        // window_.set_root_folder("ui/"); // Relative if you have ui folder next to the executable
        // window_.set_root_folder("ui"); // Relative if you have ui folder next to the executable
    }

    if (serverMode)
    {
        window_.set_port(8080);
        std::string_view url = window_.start_server("test.html");
    }
    else
    {
        window_.show("test.html");
    }

    webui::wait();

    return 0;
}

#if defined(_MSC_VER)
int APIENTRY WinMain(HINSTANCE hInst, HINSTANCE hInstPrev, PSTR cmdline, int cmdshow) { return main(); }
#endif
```

All four combinations behave consistently when the developer defines:

- **Root**: VFS via `set_file_handler(vfs)` *or* filesystem root via `set_root_folder(...)`
- **Entry**: explicit entry file via `start_server("test.html")` or `show("test.html")`

Tested combinations (server available at `http://localhost:8080`):

1. VFS + server mode (`--vfs --server`)
2. No VFS + server mode (`--server`)
3. VFS + no server mode (`--vfs`)
4. No VFS + no server mode (default)

#### Notes on Root Folder Paths

When using a filesystem root:

- Both absolute and relative paths are supported.
- A trailing slash is optional (`ui` and `ui/` behave the same).

## Rationale

Relying on index auto‑detection can hide configuration errors. Explicit entry files avoid “magic,” make behavior deterministic, and keep VFS and filesystem roots aligned.

### DIFF
--- a/examples/C++/virtual_file_system/vfs.py
+++ b/examples/C++/virtual_file_system/vfs.py
@@ -12,17 +12,8 @@
 import os
 import sys
 
-def generate_vfs_header(directory, output_header, custom_index=None):
+def generate_vfs_header(directory, output_header):
     files = []
-    index_files = {}
-
-    # Handle custom index file
-    custom_index_path = None
-    if custom_index:
-        custom_index_rel = custom_index.lstrip('/').replace('\\', '/')
-        custom_index_abs = os.path.join(directory, custom_index_rel)
-        if os.path.isfile(custom_index_abs):
-            custom_index_path = '/' + custom_index_rel
 
     # Walk through the directory and collect files
     for root, _, filenames in os.walk(directory):
@@ -32,16 +23,6 @@ def generate_vfs_header(directory, output_header, custom_index=None):
             # Ensure path starts with a slash
             relative_path = '/' + relative_path.replace('\\', '/')
             files.append((relative_path, filepath))
-
-            # Check for index files
-            if filename.startswith("index."):
-                dir_path = os.path.dirname(relative_path)
-                if dir_path not in index_files:
-                    index_files[dir_path] = relative_path
-
-    # If a custom index file is provided, override the root index
-    if custom_index_path is not None:
-        index_files["/"] = custom_index_path
 
     # Generate the C header file
     with open(output_header, 'w') as header:
@@ -75,15 +56,6 @@ def generate_vfs_header(directory, output_header, custom_index=None):
 
         header.write('static const int virtual_files_count = sizeof(virtual_files) / sizeof(virtual_files[0]);\n\n')
 
-        header.write('static const char* index_files[] = {\n')
-        for dir_path, index_path in index_files.items():
-            if dir_path == "/":
-                header.write(f'    "/", "{index_path}",\n')
-            else:
-                header.write(f'    "{dir_path}/", "{index_path}",\n')
-        header.write('    NULL\n')
-        header.write('};\n\n')
-
         header.write('bool virtual_file_system(const char* path, const unsigned char** file, int* length) {\n')
         header.write('    for (int i = 0; i < virtual_files_count; ++i) {\n')
         header.write('        if (strcmp(virtual_files[i].path, path) == 0) {\n')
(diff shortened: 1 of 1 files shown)


AlbertShown:
Thank you for PR that makes the VFS experience better and more stable.

Same missing minor parts in this PR:

1. All `vfs.py` files in `examples` folder should be updated as well
2. The version number in `vfs.py` should be incremented

However, those updates can be done in another PR.

triuk:
1. I see, they are on two more places. To avoid this redundancy, I suggest moving the script to `/example/virtual_file_system/vfs.py` and just write to use it from there in the examples' description and code comments.
2. I can do it alongside with the version change in another PR. Should it be a mayor 2.0.0 ?

AlbertShown:
1. Indeed, It is a redundancy, but I noticed some users copy an example and use it as a start-up project, so we would like to keep all necessary files in the example folder. So, let's keep those `vfs.py` files as it is.
2. Yes, you can increase it to `2.0.0`

AlbertShown:
Your Python fix is working for `http://localhost:8080` because `my_window.show("index.html")`, 
but if you tried `http://localhost:8080/subFolder` it will fail.

So, I [restored](https://github.com/webui-dev/webui/commit/724511455d9863802ee9a3877eb807a07d10a31b) old VFS **v1.1** (_Not v1.2 that has index file argument_), and I kept your C fix.

Now, old VFS v1.1 + Your C fix looks working in all scenarios.

I [updated](https://github.com/webui-dev/webui/commit/edd589a09b605ff8eef5fe72c449f042bef09ac6) all Python scripts as well.

Please let me know if there is any issue.

Thank you @triuk for the C fix.

triuk:
> Your Python fix is working for `http://localhost:8080` because `my_window.show("index.html")`, but if you tried `http://localhost:8080/subFolder` it will fail.

To clarify the use‑case: are you expecting folder URLs to resolve to a folder‑specific index file, e.g.

- `/` → `index.html`
- `/subFolder0/` → `index.html`
- `/subFolder1/` → `index.js`
- `/subFolder2/` → `index.ts`

If that’s the expectation, it can’t be reliably supported with a single entry string like `start_server("custom.html")`. Direct paths still work (`/subFolder1/index.js`), so users can navigate explicitly. But removing index autodetection means `/subFolder` won’t automatically map to `index.*`, that's true.
**The only way would be to have `custom.html` for subfolders too.** - I can try to make this solution.

If we keep index autodetection, the order should be explicit (e.g. `index.html` → `index.js` → `index.ts`). Relying on filesystem order is unstable; otherwise a file like `index.css` could be picked first depending on scan order.
 
### Summary

- If all subfolders use the same index filename, we can extend the redirect logic to cover subfolders as well.
- If subfolders use different index extensions, I suggest an explicit priority order: `index.html` → `index.js` → `index.ts` to avoid unintended picks (e.g. `index.css`). See:  https://github.com/webui-dev/webui/issues/673#issuecomment-3711735810

@AlbertShown which of these would you like to apply?

AlbertShown:
Let me think about it, because there is many scenarios:

1. User call `show("custom.html")`, No VFS
2. User call `show("custom.html")`, With a VFS
3. User only start a WebUI web server, then browser requested `/`, No VFS
4. User only start a WebUI web server, then browser requested `/`, With a VFS
5. User call `show("custom.html"`), Then requested `/subFolder0/`, No VFS
6. User call `show("custom.html")`, Then requested `/subFolder0/`, With a VFS
7. User only start a WebUI web server, then browser requested `/subFolder0/`, No VFS
8. User only start a WebUI web server, then browser requested `/subFolder0/`, With a VFS
9. Someone suggested to create a packager tool for WebUI after release, that packager may not use Python VFS at all and relay on it's own redirection mechanism... therefore, WebUI should have an option to let VFS decide what to do

triuk:
@AlbertShown
I just wanted to let you know I’m working on this. I’ll open the PR after I finish testing.

I still believe any filesystem layer should only do one thing: serve the requested file, or return `NULL` if it’s not found.

I rewrote the `_webui_external_file_handler` logic so it first tries the user-defined `custom.file`, then falls back to a hardcoded `index.*` list if needed. This now works for subfolders too.

Also, if someone wants to rely purely on the hardcoded `index.*` fallback, they can pass an empty argument, e.g. `start_server();`.

AlbertShown:
When I mapped out the 9 scenarios, I realized the file handler mechanism might need a fresh perspective, so I've added it to the to-do list for later... @triuk, if you're up for tackling this complex challenge, it would be incredibly valuable. The solution might end up being simple, but it needs to cover all 9 scenarios. I'd be thrilled to see what you come up with.

PR #684, opened 2026-02-16
add class-based call_js_from_cpp example and member-function bind overload

triuk:
Adds ergonomic class-member binding in the C++ API and introduces a class-based example.
The previous callback signature made C++ class usage too complex. 
Now you can do just this in the constructor and you are good to go:

```
my_window.bind("my_function_count", this, &CounterApp::my_function_count);
```

This update enables direct, readable binding of member functions without workarounds. Just like in the Python-WebUI.

## Changes

- `include/webui.hpp`
  - switched callback storage in `window::event::handler` to `std::function<void(event*)>` (from raw function pointers)
  - updated internal callback bookkeeping to map + mutex
  - added overload: `bind(element, instance, &Class::method)`

- `examples/C++/call_js_from_cpp_class/main.cpp`
  - new example equivalent to `call_js_from_cpp`, but implemented with a class

- `CMakeLists.txt`
  - added `call_js_from_cpp_class` target

- `examples/C++/README.md`
  - listed the new example

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,15 +122,18 @@ if (WEBUI_BUILD_EXAMPLES)
 
     add_executable(minimal ${CMAKE_CURRENT_SOURCE_DIR}/examples/C++/minimal/main.cpp)
     add_executable(call_js_from_cpp ${CMAKE_CURRENT_SOURCE_DIR}/examples/C++/call_js_from_cpp/main.cpp)
+    add_executable(call_js_from_cpp_class ${CMAKE_CURRENT_SOURCE_DIR}/examples/C++/call_js_from_cpp_class/main.cpp)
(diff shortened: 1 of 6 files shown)


AlbertShown:
Thank you @triuk for this PR.
C to C++ class method calls was in the todo list for a while.
Thank you for fixing this, very useful.

After marge, I will replace the old example `call_js_from_cpp` with this new one.

triuk:
@AlbertShown Hi, I just found `serve_a_folder` example, which uses class by default. I did not notice it as it was not a build target in CMakeFiles.txt.
I tried to rewrite it with the new class bind to have up-to-date code, but **I found a bug**. I am on it, probably solved. I will do another PR after more testing.

AlbertShown:
Sounds good @triuk, Thank you.

PR #685, opened 2026-02-17
Fix C++ event window lookup and Linux WebKit navigation handling; update serve_a_folder example

triuk:
## Summary

This PR includes three follow-up changes after the member-function bind support:

1. C++ API fix for the b36c04e:
   - `event::get_window()` now resolves the window using `bind_id` (not `window`), preventing invalid callback-window lookups.

2. Linux/WebKit navigation fix:
   - `webui_navigate()` Linux WebView path now goes through `webui_show(window, url)` to avoid navigation re-entrancy crashes when called from all-events navigation handlers.

3. Example updates:
   - `examples/C++/serve_a_folder` was refactored to a class-based style using member-function binding.
   - Added `serve_a_folder` to CMake examples target registration.
   - Updated C++ examples README accordingly.

## Validation

- Built successfully in debug build (`serve_a_folder` target).
- Verified that:
  - `OpenNewWindow` works,
  - `SwitchToSecondPage` works,
  - link navigation from the page no longer crashes in the tested Linux/WebKit setup.
- I also verified examples on Windows (Chromium)
- I did not verify on macOS.


AlbertShown:
Thank you @triuk, useful contribution

PR #691, opened 2026-03-10
Improve directory redirects and strict entry-file handling for root folders and VFS

triuk:
## Summary

This PR improves directory redirect and fallback behavior when serving UI content from either a root folder or a VFS. It also adds default arguments to the C++ `show()`, `show_browser()`, and `start_server()` APIs.

The main change is in the resolution logic for directory requests, including nested subdirectories.

---

## Core

Redirects / fallback / err404 now work consistently for any directory depth across all modes (server / vfs) combination.

**If a custom entry file is defined, for example:**

- `start_server("custom.html")`
- `show("custom.html")`
- `show_browser("custom.html", ...)`

then requests such as:

- `/`
- `/sub`
- `/sub/foo`

with or without a trailing slash, resolve to:

- `/custom.html`
- `/sub/custom.html`
- `/sub/foo/custom.html`

> Trailing-slash redirects are normalized, so paths such as `/sub/` and `/sub/foo/` do not generate double slashes in `Location`.

**If no custom filename is defined:**

- `start_server()`
- `show()`
- `show_browser()`

WebUI falls back to the hardcoded index list in this order:

1. `index.html`
2. `index.htm`
3. `index.ts`
4. `index.js`

> This also works for nested subdirectories.

**If an explicit filename does not exist, for example:**

- `start_server("not_exists.html")`
- `show("not_exists.html")`
- `show_browser("not_exists.html", ...)`

then the request remains strict and returns `Resource Not Available`.
This avoids implicit "magic" behavior and keeps the intent explicit:

- use an explicit filename when an exact file is required
- use an empty argument when fallback behavior is desired

---

### C server mode

Explicit custom entry file:

```c
(void)webui_start_server(window, "custom.html");
```

Fallback mode using the hardcoded `index.*` list:

```c
(void)webui_start_server(window, "");
```

### C normal mode

Explicit custom entry file:

```c
webui_show(window, "custom.html");
webui_show_browser(window, "custom.html", AnyBrowser);
```

Fallback mode using the hardcoded `index.*` list:

```c
webui_show(window, "");
webui_show_browser(window, "", AnyBrowser);
```

### C++ server mode

Explicit custom entry file:

```cpp
window_.start_server("custom.html");
```

Fallback mode using the hardcoded `index.*` list:

```cpp
window_.start_server();
```

### C++ normal mode

Explicit custom entry file:

```cpp
window_.show("custom.html");
window_.show_browser("custom.html");
```

Fallback mode using the hardcoded `index.*` list:

```cpp
window_.show();
window_.show_browser();
```

> In C++, `show_browser()` now defaults to `AnyBrowser` when no browser is specified.

---

## Examples

This PR adds dedicated `test_index_redirect` examples for both C and C++.
The test binaries accept `--server` and `--vfs`, so the redirect behavior can be checked across all relevant combinations.

> Do not forget to change `set_root_folder` path according to your situation when using non-vfs mode.

**This PR also updates the VFS example generators:**

- all example `vfs.py` scripts were unified to the same `v2.0` generator format
- generation of the default index list inside `vfs.py` was removed
- all corresponding `vfs.h` files were regenerated

---

## Notes

The redirect / fallback / err404 behavior was verified for:

- root-folder mode / VFS mode | server / non-server
- custom entry file present / empty argument fallback / custom entry file non-existent
- nested subdirectories
- trailing slash variants
- Linux / Windows

---


AlbertShown:
Thank you @triuk for the PR, looks like you solved the puzzle. I will re-test it again after merge, but look like we have some examples build error that prevent it from margin.

triuk:
Yes, I'll look at those automatic suggestions and build errors. I've only tested it on my Win/Lin setup and did the PR at 2 am :)
The merging probably not fails due to failed builds, but from failed "resolve" of those automatic Chatgpt codex suggestions, as I have it active for PR - I'll address it tomorrow. After that, I think it should merge.
@AlbertShown meanwhile you could check it using my fork: https://github.com/triuk/webui/tree/main

AlbertShown:
I think the build issue is due to the missing `Makefile` and `GNUmakefile` in `examples/C/test_index_redirect`

AlbertShown:
`examples/C++/test_index_redirect` as well

triuk:
I activated workflow at my fork, so I can check builds in advance next time.
When I fix those failed builds, I'll resolve the second comment and it should merge.

PR #709, opened 2026-07-24
Fix Windows WebView icon fallback to use executable icon

### LINKED ISSUE #708: Windows WebView ignores executable application icon unless it uses resource ID 101
### Description

On Windows, WebUI’s embedded WebView window does not use the application icon embedded in the executable when that icon uses the standard `IDI_APPLICATION` resource ID (32512).

The executable itself is branded correctly in Explorer and its file properties, but the WebView window, taskbar button, window caption, and Alt+Tab entry display the generic Windows application icon.

### Current behavior

The Windows WebView window class only attempts to load icon resource 101:

```c
wc.hIcon = LoadIcon(GetModuleHandle(NULL), MAKEINTRESOURCE(101));
if (!wc.hIcon)
    wc.hIcon = LoadIcon(NULL, IDI_APPLICATION);
```

When resource 101 is absent, WebUI immediately loads the generic system icon.

Many Windows build systems, including the .NET SDK, embed the application icon using `IDI_APPLICATION`, whose numeric resource ID is 32512.

### Steps to reproduce

1. Build a Windows application with a custom executable icon stored as resource 32512 and no icon resource 101.
2. Open an embedded WebView using WebUI.
3. Observe that the executable icon is correct in Explorer.
4. Observe that the WebView window and taskbar use the generic Windows icon.

### Expected behavior

WebUI should use the application icon embedded in the executable before falling back to the generic system icon.

### Proposed change

Preserve resource 101 for compatibility, then try `IDI_APPLICATION` in the executable:

```c
wc.hIcon = LoadIcon(wc.hInstance, MAKEINTRESOURCE(101));
if (!wc.hIcon)
    wc.hIcon = LoadIcon(wc.hInstance, IDI_APPLICATION);
if (!wc.hIcon)
    wc.hIcon = LoadIcon(NULL, IDI_APPLICATION);
```

With a non-null module handle, `IDI_APPLICATION` refers to resource 32512 in the executable. Passing `NULL` continues to load the generic system icon as the final fallback.

Microsoft documentation: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadiconw

ViktorJannicke:
Fixes #708

## Summary

Improve the Windows WebView window-icon fallback sequence:

1. Preserve the existing custom icon resource ID 101.
2. Fall back to the executable’s `IDI_APPLICATION` resource.
3. Fall back to the generic Windows application icon only when neither executable resource exists.

## Motivation

WebUI currently loads resource 101 and then immediately uses the generic system icon when that resource is absent.

Windows applications commonly store their executable application icon under `IDI_APPLICATION` (resource 32512). As a result, an executable can be correctly branded in Explorer while its WebUI window, taskbar button, caption, and Alt+Tab entry display the generic icon.

Using `wc.hInstance` with `IDI_APPLICATION` loads the resource from the executable. The existing `LoadIcon(NULL, IDI_APPLICATION)` call remains the final system fallback.

## Compatibility

- Existing applications using resource 101 retain precedence.
- Applications using `IDI_APPLICATION` now receive their embedded icon.
- Applications without either resource continue to receive the generic system icon.
- No public API or ABI changes are introduced.
- The change only affects the Windows embedded WebView implementation.

## Testing

Built successfully on Windows with MSVC as:

- A shared WebUI DLL
- A static WebUI library with `WEBUI_WEBVIEW_STATIC`

### DIFF
--- a/src/webui.c
+++ b/src/webui.c
@@ -12804,8 +12804,9 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpReserved) {
         wc.hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
         wc.cbClsExtra = 0;
         wc.cbWndExtra = 0;
-        wc.hIcon = LoadIcon(GetModuleHandle(NULL) ,MAKEINTRESOURCE(101));    // default user icon resouce : 101
-        if(!wc.hIcon) wc.hIcon = LoadIcon(NULL, IDI_APPLICATION);            // if not existed, use system default icon
+        wc.hIcon = LoadIcon(wc.hInstance, MAKEINTRESOURCE(101));            // default user icon resource: 101
+        if (!wc.hIcon) wc.hIcon = LoadIcon(wc.hInstance, IDI_APPLICATION);  // otherwise, use the executable's application icon
+        if (!wc.hIcon) wc.hIcon = LoadIcon(NULL, IDI_APPLICATION);          // if not present, use the system default icon
 
         if (!RegisterClassA(&wc) && GetLastError() != ERROR_CLASS_ALREADY_EXISTS) {
             _webui_wv_free(win->webView);


AlbertShown (approved):

AlbertShown:
Thank you @ViktorJannicke, much appreciated.

